PR #12, opened 2024-03-25
switch configuration types to maps

hallettj:
In some earlier feedback @dmoverton pointed out it would be better to store native queries in maps, as opposed to vecs. I thought the same logic would apply to some of the other configuration fields. It's also been bugging me that the configuration formats didn't match pattern in ddn and ndc-postgres in that we were using lists of named objects instead of maps.

I switched over some types from `Vec<StructWithNameField>` to `BTreeMap<String, StructWithoutNameField>`. To smooth things out I also added a `WithName` type that can wrap any struct, and adds a `name` field that appears at the same level as the underlying struct's fields in serialized formats. That lets us deserialize either without a name (for example if we're getting names from map keys in `schema.json` instead of from map value fields), or with a name (like if we split up schema configuration to put info for each collection into it's own file, we'll want to read the name from a field in the file.) I used that method to update `read_subdir_configs` in `directory.rs` so that it automatically grabs a name from each file.

The changes ended up being more extensive than I'd hoped so let me know if it's too much. But since this changes configuration formats I thought if we want to make this change it's better not to wait.

https://hasurahq.atlassian.net/browse/MDB-91

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1537,6 +1537,7 @@ dependencies = [
  "configuration",
  "futures-util",
  "indexmap 1.9.3",
+ "itertools 0.12.1",
  "mongodb",
  "mongodb-agent-common",
  "mongodb-support",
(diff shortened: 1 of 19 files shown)


dmoverton (approved):
Looks good to me

PR #14, opened 2024-03-26
Separate schema files for each collection

dmoverton:
## Describe your changes

Schema files will now be created and read from a subdirectory `schema/` containing one schema file per collection.
If a collection already has a schema then the `update` command skip sampling from the database for that schema.

https://hasurahq.atlassian.net/browse/MDB-83

### DIFF
--- a/crates/configuration/src/schema/mod.rs
+++ b/crates/configuration/src/schema/mod.rs
@@ -42,4 +42,23 @@ impl Schema {
             .iter()
             .map(|(name, field)| WithNameRef::named(name, field))
     }
+
+    /// Unify two schemas. Assumes that the schemas describe mutually exclusive sets of collections.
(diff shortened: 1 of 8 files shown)


crates/configuration/src/schema/mod.rs
```diff
@@ -42,4 +42,23 @@ impl Schema {
             .iter()
             .map(|(name, field)| WithNameRef::named(name, field))
     }
+
+    /// Unify two schemas. Assumes that the schemas describe mutually exclusive sets of collections.
+    pub fn merge(schema_a: Schema, schema_b: Schema) -> Schema {
```
hallettj, line 47: Would it be helpful to emit errors if there are conflicting collection or object type names? I suppose if they're auto-generated it's not likely to come up

hallettj (approved):
Looks good!

PR #16, opened 2024-03-26
arguments for read-write native queries

hallettj:
Hooks up native query arguments, but only for mutations for this PR. Converts JSON argument values to BSON according to the native query's parameter types. Implements a system to interpolate argument values into the BSON command document sent to MongoDB.

I'm going to handle arguments for read-only native queries in a follow-up PR since I realized that there is still more to do to resolve arguments in the presence of query variables.

It turns out that the earlier change I made to switch configuration fields to maps broke YAML deserialization due to a bug described here: https://github.com/dtolnay/serde-yaml/issues/344. So I've converted the native query fixtures to JSON for the time being. We'll have to either find a workaround for that bug, or disable YAML parsing. I'm inclined to do a bit of work to try for a workaround since I think the YAML format is nicer for working with: it's much less verbose, and I had to delete a comment I had in one fixture because JSON.

Ticket: https://hasurahq.atlassian.net/browse/MDB-87


crates/configuration/src/native_queries.rs
```diff
@@ -22,13 +22,46 @@ pub struct NativeQuery {
...
     pub result_type: Type,
 
-    /// Arguments for per-query customization
+    /// Arguments to be supplied for each query invocation. These will be substituted into the
+    /// given `command`.
+    ///
+    /// Argument values are standard JSON mapped from GraphQL input types, not Extended JSON.
+    /// Values will be converted to BSON according to the types specified here.
     #[serde(default)]
     pub arguments: BTreeMap<String, ObjectField>,
```
hallettj, line 31: Technically this field should be called `parameters`. Is it worth changing?

dmoverton: If `parameters` is what we use elsewhere for native queries then it's probably worth changing, given that users will need to write these.

crates/mongodb-agent-common/src/query/arguments/json_to_bson.rs
```diff
@@ -0,0 +1,424 @@
...
+            "minKey": {},
+            "maxKey": {},
+            "symbol": "a_symbol",
+        });
+
+        let expected = bson::doc! {
+            "double": Bson::Double(3.14159),
+            "int": Bson::Int32(3),
+            "long": Bson::Int64(3),
+            "decimal": Bson::Decimal128(bson::Decimal128::from_str("3.14159")?),
```
dmoverton, line 351: Clippy really doesn't like you trying to impersonate pi. I found that with one of my test cases too 😂 

https://github.com/hasura/ndc-mongodb/actions/runs/8446549353/job/23135506300?pr=16#step:6:163

crates/mongodb-agent-common/src/query/arguments/json_to_bson.rs
```diff
@@ -0,0 +1,424 @@
...
+                ),
+                ObjectField::new("minKey", Type::Scalar(BsonScalarType::MinKey)),
+                ObjectField::new("maxKey", Type::Scalar(BsonScalarType::MaxKey)),
+                ObjectField::new("symbol", Type::Scalar(BsonScalarType::Symbol)),
+            ]),
+            description: Default::default(),
+        };
+
+        let input = json!({
+            "double": 3.14159,
```
dmoverton, line 321: Clippy really doesn't like you trying to impersonate pi. I found that with one of my test cases too 😂 

https://github.com/hasura/ndc-mongodb/actions/runs/8446549353/job/23135506300?pr=16#step:6:163

dmoverton (approved):
Looks good

crates/configuration/src/native_queries.rs
hallettj: ndc-postgres uses the name `arguments` which is what I copied here. It's just that technically they're wrong: parameters are input variables, arguments are the values that bind to parameters. But I can see the argument that consistency is more important than technical correctness.

crates/mongodb-agent-common/src/query/arguments/json_to_bson.rs
hallettj: That's hilarious. And something I didn't realize is that the CI setup is running lints on test code, while it looks like when I run `cargo clippy` locally it doesn't check anything annotated with `#[cfg(test)]`

PR #17, opened 2024-03-27
Add "sensible" default behaviour for the CLI `update` command

dmoverton:
## Describe your changes

Attempt to provide some "sensible" default behaviour for the `update` command if no command line parameters are given.
The idea is to have something that will work well for most users when run from `ddn quickstart` or `ddn dev`.

- Default behaviour is now to use the validator schema if a collection has one.
- If there is no validator schema then sample documents from the collection instead.
- If no `--sample-size` is given on the command line then default to a sample size of 10.
- Add new command line flag `--no-validator-schema` which overrides the default behaviour and prevents the command from attempting to use validator schemas.

[MDB-83](https://hasurahq.atlassian.net/browse/MDB-83)

[MDB-83]: https://hasurahq.atlassian.net/browse/MDB-83?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ This changelog documents the changes between release versions.
 ## [Unreleased]
 - Use separate schema files for each collection
 - Don't sample from collections that already have a schema
+- New default behaviour for `update` CLI command: for each collection
+  - attempt to use validator schema if available
+  - if no validator schema, sample documents
+  - default sample size is 10
 
 ## [0.0.2] - 2024-03-26
 - Rename CLI plugin to ndc-mongodb ([PR #13](https://github.com/hasura/ndc-mongodb/pull/13))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,11 +3,13 @@ This changelog documents the changes between release versions.
 
 ## [Unreleased]
 - Use separate schema files for each collection
-- Don't sample from collections that already have a schema
-- New default behaviour for `update` CLI command: for each collection
-  - attempt to use validator schema if available
-  - if no validator schema, sample documents
-  - default sample size is 10
+- Changes to `update` CLI command:
+  - new default behaviour:
+    - attempt to use validator schema if available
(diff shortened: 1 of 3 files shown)


crates/cli/src/introspection/validation_schema.rs
```diff
@@ -20,42 +19,34 @@ pub async fn get_metadata_from_validation_schema(
...
+            .options
+            .validator
+            .as_ref()
+            .and_then(|x| x.get("$jsonSchema"));
+
+        match schema_bson_option {
+            Some(schema_bson) => {
+                let validator_schema =
+                    from_bson::<ValidatorSchema>(schema_bson.clone()).map_err(|err| {
+                        MongoAgentError::BadCollectionSchema(
```
dmoverton, line 38: If the collection has a something in `$jsonSchema` that can't be parsed as a `ValidatorSchema` then we treat this as an error. Perhaps we should just log this and continue, in which case we would fall back to sampling for this collection. But I think it's probably better to treat it as an error.

CHANGELOG.md
```diff
@@ -3,7 +3,13 @@ This changelog documents the changes between release versions.
...
 ## [Unreleased]
 - Use separate schema files for each collection
-- Don't sample from collections that already have a schema
+- Changes to `update` CLI command:
+  - new default behaviour:
+    - attempt to use validator schema if available
+    - if no validator schema then sample documents from the collection
+  - don't sample from collections that already have a schema
+  - if no --sample-size given on command line, default sample size is 10
+  - new option --no-validator-schema to disable attempting to use validator schema
```
daniel-chambers, line 12: Can we link to the PRs for these changelog entries?

daniel-chambers (approved):
LGTM!

CHANGELOG.md
dmoverton: Done

PR #18, opened 2024-03-27
Add `any` type

dmoverton:
## Describe your changes

Introduces a new type `any` which can be any BSON value. In NDC we represent it as a nullable scalar type called `any`, but in our internal `Type` type we keep it separate from the `BsonScalarType`s.

When unifying schemas derived from different sample documents, any type mismatches are now represented using `any` instead of resulting in an error.

[MDB-80](https://hasurahq.atlassian.net/browse/MDB-80)

[MDB-80]: https://hasurahq.atlassian.net/browse/MDB-80?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


hallettj (approved):
Looks great!

PR #20, opened 2024-03-28
Fix default config directory and port

dmoverton:
## Describe your changes

Set default config directory and port to conform to https://github.com/hasura/ndc-hub/blob/main/rfcs/0000-deployment.md

### DIFF
--- a/nix/docker-connector.nix
+++ b/nix/docker-connector.nix
@@ -8,8 +8,8 @@
 }:
 
 let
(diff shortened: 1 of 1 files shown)


daniel-chambers (approved):
This seems fine to me, but I don't really understand the Nix stuff enough to know if there might be a consequence elsewhere for doing this. I don't think so, because otherwise I'd expect to see that path mentioned elsewhere too, and it isn't.

PR #22, opened 2024-03-28
Propagate types from the v3 schema into the v2 query request

daniel-chambers:
In v2 queries, the types of columns or scalar values were often inlined into the query. In v3, the types are omitted from the query and the connector is expected to remember the types from the schema stored in its state.

This PR updates the code that maps v3 queries into v2 queries to ensure all types are inlined into the v2 query from the v3 schema.

Some bugs were fixed along the way:
* Queries with predicates that compare against columns in the root query table have been blocked as currently unsupported instead of incorrectly reading the column from the current table
* Order by elements that involved relations were not correctly generating a relations (`v2::OrderBy.relations`) map. They now are.
* v3 queries with predicates that involve comparisons against columns in other tables via the column path have been blocked as not supported. This was previously incorrectly mapped to the `v2::ComparisonColumn.path` property, but this property only ever supported the current table and the query root table, and never an arbitrary path of relationships like v3.

Also: v3 scalar types were being generated every time they were needed; however, this is static data, so they have been generated once and stored in a lazy static and are now reused everywhere.

JIRA: [MDB-7](https://hasurahq.atlassian.net/browse/MDB-7)

[MDB-7]: https://hasurahq.atlassian.net/browse/MDB-7?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


crates/mongodb-connector/src/api_type_conversions/query_request.rs
```diff
@@ -381,89 +576,121 @@ fn v3_to_v2_expression(
...
-                } => v2::ExistsInTable::UnrelatedTable {
-                    table: vec![collection],
+        } => v3_to_v2_binary_comparison(context, root_collection_object_type, object_type, column, operator, value),
+        v3::Expression::Exists { in_collection, predicate, } => {
+            let (in_table, collection_object_type) = match in_collection {
+                v3::ExistsInCollection::Related { relationship, arguments: _ } => {
+                    let v3_relationship = lookup_relationship(collection_relationships, &relationship)?;
+                    let v3_collection = context.find_collection(&v3_relationship.target_collection)?;
+                    let collection_object_type = context.find_object_type(&v3_collection.r#type)?;
+                    let in_table = v2::ExistsInTable::RelatedTable { relationship };
```
hallettj, line 586: Maybe need to capture a predicate here? But that is beyond the scope of propagating types so just something to think about following up on.

crates/mongodb-connector/src/api_type_conversions/query_request.rs
```diff
@@ -203,53 +358,90 @@ fn v3_to_v2_order_by_element(
...
+
+        let subrelations = v3_to_v2_target_path_step::<T>(context, collection_relationships, root_collection_object_type, path_iter, v2_path)?;
+        
+        v2_relations.insert(
+            path_element.relationship, 
+            v2::OrderByRelation {
+                r#where: where_expr,
+                subrelations,
+            }
+        );
```
hallettj, line 441: Ok so you're hoisting the order_by predicates. Good thinking. I forget if we still need to do something similar for comparison targets. I'm not sure if my TODO comment still applies to that case.

crates/mongodb-connector/src/main.rs
```diff
@@ -9,6 +9,9 @@ use std::error::Error;
 
 use mongo_connector::MongoConnector;
 
+#[macro_use]
+extern crate lazy_static;
```
hallettj, line 13: You don't need `macro_use` anymore. You can import macros in the modules where they are used, in this case with `use lazy_static::lazy_static`

hallettj (approved):
Thanks Daniel, this is a great help! I appreciate all the improvements you've made here - it's a reassurance to have another set of eyes looking at this

hallettj (approved):
Oh sorry - it wasn't an issue with your changes. The connector configuration fixture hadn't been updated for the schema directory configuration change.

crates/mongodb-connector/src/api_type_conversions/query_request.rs
daniel-chambers: I don't think so. In v3 there's no predicate on the `v3::ExistsInCollection::Related`, and in v2 additional predicates are already inlined into the overall predicate expression.

It's different in order by because order by isn't a predicate structure, so it needed a place to put additional non-join predicates (ie permissions predicates). Where clause predicates are a predicate structure so the permissions predicates are inlined into the overall predicate expression.

crates/mongodb-connector/src/main.rs
daniel-chambers: Ah nice one. Fixed.

crates/mongodb-connector/src/api_type_conversions/query_request.rs
```diff
@@ -41,7 +93,13 @@ pub fn v3_to_v2_query_request(
     })
 }
 
-fn v3_to_v2_query(context: &QueryContext, query: v3::Query) -> Result<v2::Query, ConversionError> {
+fn v3_to_v2_query(
+    context: &QueryContext,
+    collection_relationships: &BTreeMap<String, v3::Relationship>,
+    root_collection_object_type: &WithNameRef<schema::ObjectType>,
+    query: v3::Query,
+    collection_object_type: &WithNameRef<schema::ObjectType>,
```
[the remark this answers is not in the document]
daniel-chambers: I thought about it but if we did that we'd need to stop the caller from being the one constructing and passing the `QueryContext` and make it internal use only. Right now, this struct is used for the caller to pass contextual information, not as a way for this module to have convenient access to "global" information.

I'm going to defer doing this, because frankly this whole module will be deleted when we eventually get rid of the v2 compat layer and go v3 native.

crates/mongodb-connector/src/api_type_conversions/query_request.rs
daniel-chambers: No, with comparison targets in predicates, and predicates from the permissions system (which is what these order by predicates are for) should already been inlined into the predicate body.

daniel-chambers:
@hallettj Thanks for fixing those integration tests. 🙏

PR #26, opened 2024-04-03
Fix mapping of aggregate functions in v3->v2 query translation

daniel-chambers:
## Describe your changes
This PR fixes the incorrect mapping of aggregation functions in the v3->v2 query translation logic. The code was incorrectly trying to find the aggregation function in the native query functions data structure instead of looking up the aggregate function off of the scalar type in the NDC schema.

This broken logic was the cause of some of the failing ndc-test tests.

This PR also fixes the Nix config for running ndc-test, which was lacking a dependency from the test suite container to the MongoDB container, which caused the MongoDB container to not start when the tests started. 

The direnv config has been modified to actually activate the Nix flake, because that was the point of it in the first place. 😄


dmoverton (approved):
Looks good to me

hallettj (approved):
Good stuff, thanks for catching that!

PR #27, opened 2024-04-04
Avoid (incorrectly) deserializing already serialized v2 query response

dmoverton:
Avoid deserializing the serialized query response back into a  v2 `QueryReponse`. This was broken because the deserializer does not have enough information to differentiate between object-valued fields and relations.

The reason we were deserializing the v2 response was so that we could convert it into a v3 `QueryResponse`. Fortunately, it is not necessary to do this. The v2 and v3 query responses have almost the same JSON representation. The only difference is that in v3 the response is returned within a (singleton) array.

Therefore, rather than fully deserializing to a v2 `QueryResponse`, all we need to do is deserialize to a JSON value, wrap it in an array and serialize again.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 This changelog documents the changes between release versions.
 
 ## [Unreleased]
+- Fix bug in v2 to v3 conversion of query responses containing nested objects ([PR #27](https://github.com/hasura/ndc-mongodb/pull/27))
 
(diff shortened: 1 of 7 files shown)


hallettj (approved):
Looks good, but we'll have to rewrite this shortly to serialize each value using `bson_to_json`. If we have to do either that before converting to `Value`, or convert the extended JSON from each value back to BSON to then translate to simple JSON.

PR #32, opened 2024-04-08
Fix collection schemas not having a unique constraint for the `_id` column

daniel-chambers:
The schema endpoint was not listing a unique constraint for collections that captured the `_id` column that every Mongo table must have that is the primary key of the table.

It now generates a unique constraint named `<collection name>_id` for that column, so long as the table's object type indicates the presence of an `_id` column of type object id. That should always be the case, but if it's missing for some (bad) reason, we just omit the unique constraint too.

The test configuration in `fixtures/connector/chinook` has been updated to fix the `_id` columns being nullable, which they aren't. `_id` must always have a value.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ This changelog documents the changes between release versions.
 - Fix bug in v2 to v3 conversion of query responses containing nested objects ([PR #27](https://github.com/hasura/ndc-mongodb/pull/27))
 - Fixed bug where use of aggregate functions in queries would fail ([#26](https://github.com/hasura/ndc-mongodb/pull/26))
 - Rename Any type to ExtendedJSON to make its representation clearer ([#30](https://github.com/hasura/ndc-mongodb/pull/30))
+- The collection primary key `_id` property now has a unique constraint generated in the NDC schema for it ([#32](https://github.com/hasura/ndc-mongodb/pull/32))
 
 ## [0.0.3] - 2024-03-28
 - Use separate schema files for each collection ([PR #14](https://github.com/hasura/ndc-mongodb/pull/14))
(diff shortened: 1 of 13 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ This changelog documents the changes between release versions.
 - Fix bug in v2 to v3 conversion of query responses containing nested objects ([PR #27](https://github.com/hasura/ndc-mongodb/pull/27))
 - Fixed bug where use of aggregate functions in queries would fail ([#26](https://github.com/hasura/ndc-mongodb/pull/26))
 - Rename Any type to ExtendedJSON to make its representation clearer ([#30](https://github.com/hasura/ndc-mongodb/pull/30))
+- The collection primary key `_id` property now has a unique constraint generated in the NDC schema for it ([#32](https://github.com/hasura/ndc-mongodb/pull/32))
```
sordina, line 9: 👍

crates/mongodb-connector/src/schema.rs
```diff
@@ -86,14 +87,36 @@ fn map_type(t: &schema::Type) -> models::Type {
...
+        models::Type::Nullable { .. } => None,
+        models::Type::Array { .. } => None,
+        models::Type::Predicate { .. } => None,
+    }?;
+    let uniqueness_constraint = models::UniquenessConstraint {
+        unique_columns: vec!["_id".into()]
+    };
+    let constraint_name = format!("{}_id", name);
+    Some((constraint_name, uniqueness_constraint))
+}
```
sordina, line 108: 👍

sordina (approved):
Looks great!

hallettj (approved):
Good stuff

PR #33, opened 2024-04-11
update local services and fixtures to use latest engine version

hallettj:
Update local docker-compose configuration to use the latest engine version. This also switches to pulling engine source from the open source repo.

The implementation for dev-auth-webhook changed to a Rust project, so the nix configuration is updated accordingly.

Ticket: https://hasurahq.atlassian.net/browse/MDB-109


daniel-chambers (approved):
LGTM

PR #37, opened 2024-04-14
fix order of results for query requests with more than 10 variable sets

hallettj:
## Describe your changes

In query requests with more than 10 variable sets results were sent back in the wrong order because facet fields were put into a `BTreeMap` which sorted them lexicographically. This PR switches to a `Vec` of tuples to preserve expected order.

While working on this I updated the dev services to run two connectors with a remote relationship configured. If you run `arion up -d` you can access GraphiQL at http://localhost:7100/ and test the fix with a query like this one:

```graphql
query AlbumMovies {
  album(limit: 11) {
    title
    movies(limit: 2) {
      title
      runtime
    }
    albumId
  }
}
```

The remote relationship matches `albumId` in the `Album` collection of the chinook database with the `runtime` field in the `movies` collection of the sample_mflix database. It's not a sensical relationship but it gives us something to test with since both of those fields are ints with plenty of values that line up.

[MDB-111](https://hasurahq.atlassian.net/browse/MDB-111)

[MDB-111]: https://hasurahq.atlassian.net/browse/MDB-111?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


crates/mongodb-agent-common/src/query/foreach.rs
```diff
@@ -56,7 +56,7 @@ pub fn pipeline_for_foreach(
     foreach: Vec<ForeachVariant>,
     query_request: &QueryRequest,
 ) -> Result<(Pipeline, ResponseShape), MongoAgentError> {
-    let pipelines_with_response_shapes: BTreeMap<String, (Pipeline, ResponseShape)> = foreach
+    let pipelines_with_response_shapes: Vec<(String, (Pipeline, ResponseShape))> = foreach
```
hallettj, line 59: This PR contains a lot of fixture and test changes. This line is the only change required for the actual fix.

crates/mongodb-connector/src/mongo_connector.rs
```diff
@@ -160,8 +161,6 @@ impl Connector for MongoConnector {
         let response = handle_query_request(state, v2_request)
             .await
             .map_err(mongo_agent_error_to_query_error)?;
-        let r = v2_to_v3_query_response(response);
-        tracing::warn!(v3_response = %serde_json::to_string(&r).unwrap());
```
hallettj, line 164: This fixes some debugging output I accidentally committed in a previous PR. I'd like to set up the sdk to making the logging level for this crate configurable so that I don't have to change `debug!`s to `warn!`s to see these traces.

dmoverton (approved):

PR #39, opened 2024-04-14
switch to query error variant that propagates error messages to GraphQL responses

hallettj:
## Describe your changes

When we encounter problems with incoming query requests we've been sending back `QueryError::InvalidQuery` error values. But when the engine receives these it produces GraphQL responses that only say "internal error". This change switches to `QueryError::UnprocessableContent`. With this change the engine copies error messages from the connector to the GraphQL response.


dmoverton (approved):

hallettj:
When we encounter problems with incoming query requests we've been sending back `QueryError::InvalidQuery` error values. But when the engine receives these it produces GraphQL responses that only say "internal error". This change switches to `QueryError::UnprocessableContent`. With this change the engine copies error messages from the connector to the GraphQL response.

PR #43, opened 2024-04-17
ndc-test nix improvements and Chinook data fixes

daniel-chambers:
This PR improves some of the testing infrastructure, but stops short of actually integrating ndc-test in CI. We still need to make improvements to ndc-test and fix more bugs in the mongo connector before that will be possible.

### Nix
The Nix arion testing configuration was modified slightly to support mounting the `snapshots` directory in the repo and reading and writing ndc-test snapshots to/from that directory. There are some example commands of how to do this commented out in `project-ndc-test.nix`. However, it's still configured to just do a basic "test" run because:

- snapshots currently just create broken tests; for example, ordering tests are generated using unstable orderings, so the snapshots always fail
- snapshots generated in an unusable fashion, where they are just sprayed into the folder under folders named after the hash of the requests. This is very unhelpful, but will require improvements to ndc-test to change.

### Chinook data
The Chinook dataset had some significant issues, which have now been resolved! 😄 Namely:

* Many tables failed to import correctly from CSV due to type issues. This has been resolved by changing the import process so that it imports from JSON files instead of CSV. The JSON files were manually exported after manually massaging the CSV data into Mongo correctly and then exporting it as JSON.
* Importing from CSV every time meant that new object IDs were generated each time the import was performed. This is incompatible with snapshot testing, so now that we're importing from JSON, the Object IDs have been captured and don't change on every import.
* Some of the data types in the schema were incorrect, namely any column that dealt with money was being typed as a float, when it really should have been a decimal. The Employee.ReportsTo column was also a string, when it should have been an int. These issues have also been corrected.

JIRA: [MDB-110](https://hasurahq.atlassian.net/browse/MDB-110)

[MDB-110]: https://hasurahq.atlassian.net/browse/MDB-110?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


hallettj (approved):
Looks great, thanks for all of these improvements! We still need to update the fixture ddn to close the loop. I don't think that's a blocker for the PR since technically everything is working as is. I may get the CLI running for real to work on that.

PR #44, opened 2024-04-17
Be stricter when converting user-supplied JSON values to BSON

dmoverton:
## Describe your changes

If we don't know the correct BSON scalar type for a user-supplied value then refuse to convert it from JSON to BSON. This prevents the user inserting arbitrary values where a scalar value is expected.

[MDB-112](https://hasurahq.atlassian.net/browse/MDB-112)

### Type
- [ ] highlight
- [ ] enhancement
- [ ] bugfix
- [ ] behaviour-change
- [ ] performance-enhancement
- [x] security-fix

[MDB-112]: https://hasurahq.atlassian.net/browse/MDB-112?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


hallettj (approved):
Looks good!

PR #45, opened 2024-04-18
add native queries, functions or virtual collections defined by pipelines

hallettj:
## Describe your changes

Implements native queries. They are defined as aggregation pipelines. If the root target of a query request is a native query then the given pipeline will form the start of the overall query plan pipeline. In this case the query will be executed as a MongoDB `aggregate` command with no target collection - as opposed to our other queries which are an `aggregate` command that *does* have a collection target.

Native queries currently cannot be the target of a relation.

There is a really basic native query in fixtures to test with. If you run services with `arion up -d` you can see it as a query field called `hello`.

The changes were going to result in a large-ish amount of very similar, but technically incompatible code involving converting configuration to ndc types for schema responses, and for processing query requests. To avoid that I pushed configuration processing into the `configuration` crate. This makes it easier to share that logic, pushes a bunch of errors from connector runtime to configuration parsing time, and pushes computation to connector startup time instead of response-handling time. This resulted in a bunch of changes:

- The `MongoConfig` type is gone. Instead configuration-related data is kept in `Configuration` which is now passed directly to `mongodb-agent-common` functions. Database-connection data is put in a new type, `ConnectorState`. So now we have properly separated configuration and state types, which is what the ndc-sdk API expects.
- The `configuration` crate has a new dependency on `ndc-models`. We need to keep the `ndc-models` version matched with `ndc-sdk`. To make that easier I moved configuration for those dependencies to the workspace `Cargo.toml` and added a note.

[MDB-103](https://hasurahq.atlassian.net/browse/MDB-103)

[MDB-103]: https://hasurahq.atlassian.net/browse/MDB-103?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


crates/configuration/src/configuration.rs
```diff
@@ -1,85 +1,339 @@
...
+
+    /// Functions are based on native queries using [NativeQueryRepresentation::Function]
+    /// representation.
+    pub functions: BTreeMap<String, ndc::FunctionInfo>,
+
+    /// In query requests functions and collections are treated as the same, but in schema
+    /// responses they are separate concepts. So we want a set of [CollectionInfo] values for
+    /// functions for query processing, and we want it separate from `collections` for the schema
+    /// response.
+    pub function_collection_infos: BTreeMap<String, ndc::CollectionInfo>,
```
dmoverton, line 29: Will `functions` and `function_collection_infos` have the same set of keys?
If so, would it be better to represent this as a `BTreeMap<String, (ndc::FunctionInfo, ndc::CollectionInfo)>`?

crates/configuration/src/configuration.rs
```diff
@@ -1,85 +1,339 @@
...
+
+    /// Procedures are based on native procedures.
+    pub procedures: BTreeMap<String, ndc::ProcedureInfo>,
 
     /// Native procedures allow arbitrary MongoDB commands where types of results are
     /// specified via user configuration.
-    #[serde(default, skip_serializing_if = "BTreeMap::is_empty")]
     pub native_procedures: BTreeMap<String, NativeProcedure>,
```
dmoverton, line 36: Likewise, could `procedures` and `native_procedures` be represented as a single map `BTreeMap<String, (ndc::ProcedureInfo, NativeProcedure)>`?

crates/configuration/src/configuration.rs
```diff
@@ -1,85 +1,339 @@
...
+    let object_types_from_native_procedures = native_procedures
+        .values()
+        .flat_map(|native_procedure| &native_procedure.object_types);
+    let object_types_from_native_queries = native_queries
+        .values()
+        .flat_map(|native_query| &native_query.object_types);
+    object_types_from_schema
+        .chain(object_types_from_native_procedures)
+        .chain(object_types_from_native_queries)
+    // .map(|(name, ot)| (name.to_owned(), ot.clone()))
```
dmoverton, line 199: commented out code

crates/configuration/src/serialized/native_procedure.rs
```diff
@@ -0,0 +1,82 @@
...
+///
+/// Native Procedures appear as "procedures" in your data graph.
+#[derive(Clone, Debug, Deserialize, JsonSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct NativeProcedure {
+    /// You may define object types here to reference in `result_type`. Any types defined here will
+    /// be merged with the definitions in `schema.json`. This allows you to maintain hand-written
+    /// types for native procedures without having to edit a generated `schema.json` file.
+    #[serde(default, skip_serializing_if = "BTreeMap::is_empty")]
+    pub object_types: BTreeMap<String, ObjectType>,
```
dmoverton, line 20: I'm wondering whether all user-defined object types should be in one place, rather than each native procedure/query having their own. That would make it easier to re-use types between procedures/queries. It would also provide a place where users could override type definitions inferred by the schema inference.

crates/configuration/src/configuration.rs
hallettj: That does make sense. But I think it's a matter of style. There isn't any case where we want to look up both `FunctionInfo` and `CollectionInfo` - the schema handler wants one, the query handler wants the other. We shouldn't be constructing `Configuration` values without going through `validate` except in tests so I'm not concerned about accidentally created an invalid state.

crates/configuration/src/configuration.rs
hallettj: Yep that makes sense. But again I think it's a matter of style since there are no cases where we want both values together.

crates/configuration/src/serialized/native_procedure.rs
hallettj: Yeah I think you make a good point. I think the way I have it set up has a convenience advantage because you can define single-use types in the same place they're used. But it might be more confusing. @codedmart already ran into this trying to write a procedure that reuses a collection object type.

I think this is a good topic for discussion. I think it's outside the scope of this PR since I already established this convention in native procedures. But there's still time to change it in another ticket.

crates/configuration/src/configuration.rs
hallettj: Ok I made this change. It does make the validate function simpler. But I'd rather not combine the procedure maps since I like keeping the ndc stuff somewhat separated from the mongodb-specific stuff.

codedmart (approved):

PR #46, opened 2024-04-19
Add collection name as an option for native queries

codedmart:
## Describe your changes
Add optional `collection_name` to native query configuration. When a `collection_name` is used in configuration the pipeline is run using that collection as the base.

### Type
- [ ] highlight
- [x] enhancement
- [ ] bugfix
- [ ] behaviour-change
- [ ] performance-enhancement
- [ ] security-fix

### Changelog entry

Add optional `collection_name` to native query configuration. When a `collection_name` is used in configuration the pipeline is run using that collection as the base.

### DIFF
--- a/crates/mongodb-agent-common/src/query/execute_query_request.rs
+++ b/crates/mongodb-agent-common/src/query/execute_query_request.rs
@@ -39,8 +39,14 @@ pub async fn execute_query_request(
             let collection = database.collection(&collection_name);
             collect_from_cursor(collection.aggregate(pipeline, None).await?).await
         }
-        QueryTarget::NativeQuery { .. } => {
-            collect_from_cursor(database.aggregate(pipeline, None).await?).await
+        QueryTarget::NativeQuery { native_query, .. } => {
+            match native_query.clone().collection_name {
+                Some(coll_name) => {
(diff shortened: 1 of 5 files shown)


crates/mongodb-agent-common/src/query/execute_query_request.rs
```diff
@@ -39,8 +39,14 @@ pub async fn execute_query_request(
             let collection = database.collection(&collection_name);
             collect_from_cursor(collection.aggregate(pipeline, None).await?).await
         }
-        QueryTarget::NativeQuery { .. } => {
-            collect_from_cursor(database.aggregate(pipeline, None).await?).await
+        QueryTarget::NativeQuery { native_query, .. } => {
+            match native_query.clone().collection_name {
+                Some(coll_name) => {
+                    let collection = database.collection(&coll_name.to_string());
```
hallettj, line 45: I think you can simplify this

```suggestion
                    let collection = database.collection(coll_name);
```

hallettj (approved):
We should make a corresponding change to the explain handler - it reads `QueryTarget` and applies basically the same logic that the query handler does.

As I said, I'd like to change the name to something other than "collection_name".

Otherwise it looks good!

PR #47, opened 2024-04-20
add integration tests

hallettj:
## Describe your changes

Run integration tests with,

```sh
just test-integration
```

Or if you don't have your nix develop shell set up you can run,

```sh
nix develop -c just test-integration
```


daniel-chambers:
Just a comment on testing strategy:

While I like the idea of these sorts of end to end integration tests, I think our goal should be to have a small number of these for smoke testing and regression testing, with the majority of our tests done at the NDC level (ie. issue NDC requests and assert against NDC responses) using `ndc-test`. 

This decouples our tests from the engine, which is useful when developing features that the engine doesn't yet support, simplifies the tests a lot (no metadata required), and also isolates our tests from breaking changes in the engine (less important, since this is not supposed to happen much). It also focuses our testing investment into `ndc-test` which can be used by any connector to implement snapshot-style testing. We have failed to do this so far, with every connector either not doing testing or rolling their own testing setup. It would be good to see investment in testing be share-able across connectors, though I know that will take some upfront time and investment. cc @codedmart

codedmart (approved):

PR #48, opened 2024-04-22
run integration tests against multiple mongodb versions

hallettj:
## Describe your changes

Runs integration tests against MongoDB versions 4, 6, and 7. (The oldest version supported by MongoDB is 4.4. The most recent stable release is 7.0) Version 5 is skipped for now because there is a problem with the `hello` native query example in that version.

We can't claim to support v3.6 because the `$replaceWith` stage wasn't implemented in that version.


dmoverton (approved):

codedmart:
Curious why version 5 is having issues?

PR #50, opened 2024-04-23
Mention MONGODB_DATABASE_URI environment variable in error message if it is missing

dmoverton:
## Describe your changes

If the database URI is missing we want the error message to mention the environment variable `MONGODB_DATABASE_URI` instead of the command line option `--connection-uri`.
When run via the `ddn` CLI, the environment variable will be the primary means of specifying this option and it is confusing not to mention it.

[MDB-119](https://hasurahq.atlassian.net/browse/MDB-119)

### Type
- [ ] highlight
- [x] enhancement
- [ ] bugfix
- [ ] behaviour-change
- [ ] performance-enhancement
- [ ] security-fix

[MDB-119]: https://hasurahq.atlassian.net/browse/MDB-119?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ This changelog documents the changes between release versions.
 ## [Unreleased]
 - Fix incorrect order of results for query requests with more than 10 variable sets (#37)
 - In the CLI update command, don't overwrite schema files that haven't changed ([#49](https://github.com/hasura/ndc-mongodb/pull/49/files))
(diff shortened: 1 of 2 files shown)


daniel-chambers (approved):
I don't love how now the CLI args say `--connection-uri` is optional when it isn't. But I don't have a better solution since you can't customise the error message.

PR #53, opened 2024-04-25
serialize query responses according to schema types

hallettj:
## Describe your changes

This changes response serialization to use simple JSON instead of Extended JSON (except in cases of fields with the type `ExtendedJSON`).

Mutation responses have not been updated with this change yet - that will be in a follow-up PR.

This could be greatly simplified if we had an internal query request type that used copies of object type definitions in all places instead of object type names. I've got some thoughts on changing `v3_to_v2_query_request` to work with an internal type like that instead of with the v2 query request type. But that's a big change that will have to wait until later.

[MDB-96](https://hasurahq.atlassian.net/browse/MDB-96)

[MDB-96]: https://hasurahq.atlassian.net/browse/MDB-96?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


crates/dc-api-types/src/query.rs
```diff
@@ -16,49 +16,27 @@ pub struct Query {
     #[serde(
         rename = "aggregates",
         default,
-        with = "::serde_with::rust::double_option",
         skip_serializing_if = "Option::is_none"
     )]
-    pub aggregates: Option<Option<::std::collections::HashMap<String, crate::Aggregate>>>,
+    pub aggregates: Option<::std::collections::HashMap<String, crate::Aggregate>>,
```
hallettj, line 21: Flattening these double-option wrappings was not strictly necessary. But the double wrapping doesn't do anything for us, and I got tired of dealing with it.

crates/integration-tests/src/tests/snapshots/integration_tests__tests__basic__runs_a_query.snap
```diff
@@ -5,43 +5,53 @@ expression: "query(r#\"\n                query Movies {\n                  movie
...
     - imdb:
-        rating: 7.4
+        rating:
+          $numberDouble: "7.4"
         votes: 9847
       title: The Great Train Robbery
     - imdb:
-        rating: 7.1
+        rating:
+          $numberDouble: "7.1"
```
hallettj, line 19: The strict Extended JSON serialization for `rating` is expected. In the sample database ratings are a mixture of `double` and `int` values. Because we have two different types in the same position type unification resolves the type to `ExtendedJSON`

crates/mongodb-connector/src/api_type_conversions/query_request.rs
```diff
@@ -1269,222 +1261,4 @@ mod tests {
         assert_eq!(v2_request, expected);
         Ok(())
     }
-
-    fn make_scalar_types() -> BTreeMap<String, ScalarType> {
```
hallettj, line 1273: I moved these into a shared `test_helpers` module

hallettj:
Ready to review!

I realized that I could probably improve performance substantially by hoisting logic that's currently done per result row.

Aggregates are still serialized to extjson, and aggregates from relations aren't serialized. I'd like to fix that soon, but we don't need aggregates to work until they're supported by the engine.

codedmart:
> I realized that I could probably improve performance substantially by hoisting logic that's currently done per result row.

This is follow up work? Or you did this in this PR?

crates/configuration/src/schema/mod.rs
```diff
@@ -95,6 +95,30 @@ impl From<Type> for ndc_models::Type {
     }
 }
 
+// Should only be used for testing
```
codedmart, line 98: Do we have any way of ensuring this for all these?

crates/mongodb-connector/src/api_type_conversions/query_request.rs
```diff
@@ -105,6 +105,7 @@ fn find_object_field<'a>(
         ConversionError::UnknownObjectTypeField {
             object_type: object_type.name.to_string(),
             field_name: field_name.to_string(),
+            path: Default::default(), // TODO: set a path for more helpful error reporting
```
codedmart, line 108: Is this TODO for later or missed in this PR?

crates/mongodb-connector/src/query_response.rs
```diff
@@ -0,0 +1,815 @@
...
+    query_aggregates: &IndexMap<String, Aggregate>,
+    mut aggregate_values: BTreeMap<String, Bson>,
+) -> Result<IndexMap<String, serde_json::Value>> {
+    query_aggregates
+        .iter()
+        .map(
+            |(key, _aggregate_definition)| match aggregate_values.remove_entry(key) {
+                Some((owned_key, value)) => Ok((
+                    owned_key,
+                    // TODO: bson_to_json
```
codedmart, line 184: Also later TODO or for this PR?

codedmart (approved):
Left a few comments. Thanks!

crates/configuration/src/schema/mod.rs
hallettj: I removed these

crates/mongodb-connector/src/api_type_conversions/query_request.rs
hallettj: Saving this for later

crates/mongodb-connector/src/query_response.rs
hallettj: Filed https://hasurahq.atlassian.net/browse/MDB-130 to follow up on this

PR #58, opened 2024-04-26
Add /query trace spans and update SDK

codedmart:
## Describe your changes

Add tracing spans and update to the latest SDK

[MDB-131](https://hasurahq.atlassian.net/browse/MDB-131)

[MDB-131]: https://hasurahq.atlassian.net/browse/MDB-131?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/crates/mongodb-connector/src/mongo_connector.rs
+++ b/crates/mongodb-connector/src/mongo_connector.rs
@@ -18,7 +18,7 @@ use ndc_sdk::{
         QueryResponse, SchemaResponse,
     },
 };
-use tracing::instrument;
+use tracing::{instrument, Instrument};
 
 use crate::{
     api_type_conversions::{v2_to_v3_explain_response, v3_to_v2_query_request},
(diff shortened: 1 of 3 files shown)


crates/mongodb-connector/src/mongo_connector.rs
```diff
@@ -141,18 +141,27 @@ impl Connector for MongoConnector {
...
+            tracing::info_span!("Serialize Query Response", internal.visibility = "user").in_scope(|| {
+                serialize_query_response(&query_context, &request, response_documents)
+                .map_err(|err| {
+                    QueryError::UnprocessableContent(format!(
+                        "error converting MongoDB response to JSON: {err}"
+                    ))
+                })
+            })
+        }
+        .instrument(tracing::info_span!("/query", internal.visibility = "user"))
```
hallettj, line 163: Since this instrumentation applies to the effectively the entire function, I think another way to do this would be to modify the `#[instrument]` annotation on the function. Currently it's configured to only emit traces on errors, but it can be configured to do other things. One thing to watch out for is that using `#[instrument]` on a function captures argument values unless you opt out - the `skip_all` option excludes all arguments from the trace.

But this way works just as well.

hallettj (approved):
Looks good!

PR #68, opened 2024-04-30
Add root config and update cli default

codedmart:
## Describe your changes

Add an optional root `configuration.json` or `configuration.yaml` file to allow editing cli options. Update default sample size to 100.

[MDB-133](https://hasurahq.atlassian.net/browse/MDB-133)


crates/configuration/src/directory.rs
```diff
@@ -9,11 +9,13 @@ use std::{
 use tokio::fs;
 use tokio_stream::wrappers::ReadDirStream;
 
-use crate::{serialized::Schema, with_name::WithName, Configuration};
+use crate::{configuration::ConfigurationOptions, serialized::Schema, with_name::WithName, Configuration};
 
 pub const SCHEMA_DIRNAME: &str = "schema";
 pub const NATIVE_PROCEDURES_DIRNAME: &str = "native_procedures";
 pub const NATIVE_QUERIES_DIRNAME: &str = "native_queries";
+pub const CONFIURATION_OPTIONS_JSON: &str = "configuration.json";
```
dmoverton, line 17:
```suggestion
pub const CONFIGURATION_OPTIONS_JSON: &str = "configuration.json";
```

crates/configuration/src/directory.rs
```diff
@@ -108,6 +113,20 @@ where
     }
 }
 
+pub async fn parse_configuration_options_file(dir: &Path) -> ConfigurationOptions {
+    let json_config_file = parse_config_file(&dir.join(CONFIURATION_OPTIONS_JSON), JSON).await;
```
dmoverton, line 117:
```suggestion
    let json_config_file = parse_config_file(&dir.join(CONFIGURATION_OPTIONS_JSON), JSON).await;
```

crates/configuration/src/directory.rs
```diff
@@ -108,6 +113,20 @@ where
...
+    if let Ok(config_options) = json_config_file {
+        return config_options
+    }
+
+    let yaml_config_file = parse_config_file(&dir.join(CONFIURATION_OPTIONS_YAML), YAML).await;
```
dmoverton, line 122:
```suggestion
    let yaml_config_file = parse_config_file(&dir.join(CONFIGURATION_OPTIONS_YAML), YAML).await;
```

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 This changelog documents the changes between release versions.
 
 ## [Unreleased]
+- Add an optional root `configuration.json` or `configuration.yaml` file to allow editing cli options. Update default sample size to 100. ([#68](https://github.com/hasura/ndc-mongodb/pull/68))
```
dmoverton, line 5: I would list this as two separate changes (with the same PR reference)

dmoverton:
I think it would be useful if the `update` command wrote out a `configuration.yaml` file with the default options if it doesn't already exist. This would make it easier for the user to edit the options, rather than having to write the file from scratch. It also means users won't get an unexpected change in behaviour if the default options change in the future.

codedmart:
> I think it would be useful if the `update` command wrote out a `configuration.yaml` file with the default options if it doesn't already exist. This would make it easier for the user to edit the options, rather than having to write the file from scratch. It also means users won't get an unexpected change in behaviour if the default options change in the future.

Ah right I meant to do that and got side tracked. Thanks!

crates/configuration/src/configuration.rs
```diff
@@ -4,13 +4,32 @@ use anyhow::{anyhow, ensure};
...
+use serde::Deserialize;
 
 use crate::{
     native_procedure::NativeProcedure,
     native_query::{NativeQuery, NativeQueryRepresentation},
     read_directory, schema, serialized,
 };
 
+#[derive(Clone, Debug, Deserialize)]
+pub struct ConfigurationOptions {
```
dmoverton, line 16: Do we want some more structure in this so we can add extra sections in the future? The current options could go under an `introspection_options` key, that would leave from for other options in the future, such as for DB connection settings. E.g.
```yaml
introspection_options:
    sample_size: 100
    no_validator_schema: false
connection_options:
    whatever
...
```

crates/cli/src/introspection/type_unification.rs
```diff
@@ -17,8 +17,8 @@ type ObjectType = WithName<schema::ObjectType>;
 /// Unify two types.
 /// This is computing the join (or least upper bound) of the two types in a lattice
 /// where `ExtendedJSON` is the Top element, Scalar(Undefined) is the Bottom element, and Nullable(T) >= T for all T.
-pub fn unify_type(type_a: Type, type_b: Type) -> Type {
-    let result_type = match (type_a, type_b) {
+pub fn unify_type(type_a: Type, type_b: Type, all_schema_nullable: bool) -> Type {
```
dmoverton, line 20: `unify_type` should not need to know about `all_schema_nullable` option. This should be handled in `make_field_type` by wrapping the resulting type in `Nullable`.

crates/configuration/src/configuration.rs
```diff
@@ -167,6 +172,34 @@ impl Configuration {
...
 }
 
+#[derive(Copy, Clone, Debug, Default, Deserialize, Serialize)]
+pub struct ConfigurationOptions {
+    // Options for introspection
+    pub introspection_options: ConfigurationIntrospectionOptions,
+}
+
+#[derive(Copy, Clone, Debug, Deserialize, Serialize)]
+pub struct ConfigurationIntrospectionOptions {
```
hallettj, line 182: I hate to nitpick, but in other places where we map struct keys to config files we set `#[serde(renameAll = "camelCase")]`

hallettj (approved):

crates/configuration/src/configuration.rs
codedmart: Please nitpick. Consistency is good 😄.

PR #69, opened 2024-04-30
test query request with one variable set

hallettj:
## Describe your changes

Add a test just to make sure this case works. I thought I remembered MongoDB rejecting `$facet` stages with only one sub-pipeline, but according to this test that case works just fine in MongoDB versions 5, 6, and 7.

This PR expands the integration tests to allow sending a query request directly to a connector, bypassing the engine. It also cleans up some of the test helper code. The test snapshots all needed to be updated because I changed a function name, which changed the expression used to produce snapshot results, and the expression is captured as part of the snapshot file.

[MDB-135](https://hasurahq.atlassian.net/browse/MDB-135)

[MDB-135]: https://hasurahq.atlassian.net/browse/MDB-135?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/crates/integration-tests/src/tests/remote_relationship.rs
+++ b/crates/integration-tests/src/tests/remote_relationship.rs
@@ -29,7 +29,7 @@ async fn provides_source_and_target_for_remote_relationship() -> anyhow::Result<
 
 #[tokio::test]
 async fn handles_request_with_single_variable_set() -> anyhow::Result<()> {
-    assert_yaml_snapshot!({
+    assert_yaml_snapshot!(
         run_connector_query(
             query_request()
                 .collection("movies")
@@ -41,6 +41,6 @@ async fn handles_request_with_single_variable_set() -> anyhow::Result<()> {
                 ),
         )
         .await?
-    });
+    );
     Ok(())
 }
--- a/crates/integration-tests/src/tests/remote_relationship.rs
+++ b/crates/integration-tests/src/tests/remote_relationship.rs
@@ -1,11 +1,12 @@
-use crate::query;
+use crate::{graphql_query, run_connector_query};
 use insta::assert_yaml_snapshot;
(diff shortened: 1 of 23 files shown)


crates/integration-tests/src/tests/remote_relationship.rs
```diff
@@ -25,3 +26,21 @@ async fn provides_source_and_target_for_remote_relationship() -> anyhow::Result<
...
+                .query(
+                    query()
+                        .predicate(equal(target!("_id"), variable!(id)))
+                        .fields([field!("title")]),
+                ),
+        )
+        .await?
+    });
+    Ok(())
+}
```
hallettj, line 46: This is the new test

crates/integration-tests/src/tests/snapshots/integration_tests__tests__remote_relationship__handles_request_with_single_variable_set.snap
```diff
@@ -0,0 +1,6 @@
+---
+source: crates/integration-tests/src/tests/remote_relationship.rs
+expression: "{\n    run_connector_query(query_request().collection(\"movies\").variables([vec![(\"id\",\n                                    json!(\"573a1390f29313caabcd50e5\"))]]).query(query().predicate(equal(target!(\"_id\"),\n                                variable!(id))).fields([field!(\"title\")]))).await?\n}"
+---
+- rows:
+    - title: Gertie the Dinosaur
```
hallettj, line 6: This is the snapshot of the result of the new test

dmoverton (approved):

codedmart (approved):

PR #71, opened 2024-05-25
refactor: remove v2 types

hallettj:
Removes v2 types, and replaces them with a new set of internal types that closely match ndc models. The root of the new types is called `QueryPlan` instead of `QueryRequest`. It is a denormalized version of a query request. In particular `QuerPlan`

- puts type annotations on fields, variable references, etc.
- uses a `Type` type that inlines object type references
- collects relationship reference details from all fields, filters, and sorts and places relevant details in one  map in each sub-query (instead of having one relations map at the `QueryRequest` level, the new types have a map on each `Query` value for relations with that query's collection as a source, and that are referenced in that query)
- collects one map of join details at the `QueryPlan` level (the analog of `QueryRequest`) for existence checks on unrelated collections

The changes simplify emitting MongoDB aggregation code, and lower the impedence mismatch to using features of v3 that aren't reflected in v2.

I preserved and updated all the existing tests except one that tests filtering by fields of a nested object of a related collection which is something that worked in v2, but will require an update to the latest ndc-spec to work in v3.

There are some more changes necessary for filtering and sorting by fields of a related collection. We need to adjust the way query request expressions map to MongoDB expressions. I thought it would be best to get this refactor in, and put those changes in a separate PR.


crates/mongodb-agent-common/src/comparison_function.rs
```diff
@@ -22,11 +21,10 @@ pub enum ComparisonFunction {
     IRegex,
 }
 
-use BinaryComparisonOperator as B;
+use ndc_query_plan::QueryPlanError;
+// use BinaryComparisonOperator as B;
```
dmoverton, line 25: Remove commented out code
```suggestion
```

crates/mongodb-agent-common/src/interface_types/mongo_agent_error.rs
```diff
@@ -21,13 +20,14 @@ pub enum MongoAgentError {
     MongoDBSerialization(#[from] mongodb::bson::ser::Error),
     MongoDBSupport(#[from] mongodb_support::error::Error),
     NotImplemented(&'static str),
-    MutationError(#[from] MutationError),
+    ProcedureError(#[from] ProcedureError),
```
dmoverton, line 23: It should probably just be `Procedure` to be consistent with the other enum values
```suggestion
    Procedure(#[from] ProcedureError),
```

crates/mongodb-agent-common/src/scalar_types_capabilities.rs
```diff
@@ -64,24 +149,24 @@ pub fn comparison_operators(
...
-        aggregate_functions: Some(aggregations),
-        comparison_operators: if comparisons.is_empty() {
-            None
-        } else {
-            Some(comparisons)
-        },
-        update_column_operators: None,
-    }
-}
+// fn capabilities(scalar_type: BsonScalarType) -> ScalarTypeCapabilities {
```
dmoverton, line 152: Commented out code

dmoverton (approved):
Looks good. This is a big improvement. Just a few minor suggested changes.

It might be useful to move the `ndc-query-plan` crate to the SDK at some point so it can be used by other connectors.

hallettj:
> Looks good. This is a big improvement. Just a few minor suggested changes.
> 
> It might be useful to move the `ndc-query-plan` crate to the SDK at some point so it can be used by other connectors.

Thanks! I had the same thought.

PR #73, opened 2024-06-06
Update ndc-models and ndc-sdk to v0.1.4

dmoverton:
## Describe your changes

Update to NDC v0.1.4 and path through `file_path` properties for filtering and ordering.

### Type
- [ ] highlight
- [x] enhancement
- [ ] bugfix
- [ ] behaviour-change
- [ ] performance-enhancement
- [ ] security-fix


crates/ndc-query-plan/src/plan_for_query_request/mod.rs
```diff
@@ -137,10 +137,16 @@ fn plan_for_aggregate<T: QueryContext>(
...
+        ndc::Aggregate::ColumnCount {
+            column,
+            distinct,
+            field_path: _,
+        } => Ok(plan::Aggregate::ColumnCount { column, distinct }),
+        ndc::Aggregate::SingleColumn {
+            column,
+            function,
+            field_path: _,
+        } => {
```
hallettj, line 149: We want to forward field paths here. But we need to do a pass of testing and cleaning up aggregate support anyway so that can wait a bit.

hallettj (approved):
Looks good!

PR #74, opened 2024-06-10
handle column references that involve variables

hallettj:
To support root column references we need to handle column references that involve variables. Currently we're using the built-in variable `$$ROOT` to reference the root collection (which is wrong, but we'll fix that in another PR, and the correct solution will also require a variable.) Before this PR our column references don't work with variable references like `$$ROOT` due to the two different kinds of expressions that are found in MongoDB aggregation pipelines. Specifically there are,

- match queries, where the reference is a key in the document used in a `$match` aggregation stage
- aggregation expressions which appear in a number of contexts

The code we had would create a match stage like this:

```json
{
  "$match": {
    "$$ROOT.field": { "$eq": 1 }
  }
}
```

That doesn't work because `"$$ROOT.field"` is in a match query context which does not allow using `$` to reference field names or variables. But there is a match query operator, `$expr`, that switches to an aggregation expression context. So the correct solution looks like this:

```json
{
  "$match": {
    "$expr": {
      "$eq": ["$$ROOT.field", 1]
    }
  }
}
```

This PR updates the code for producing `$match` stages to use switch to aggregation expression context to handle cases like this. Specifically I introduced an enum, `ColumnRef`, which signals cases where the reference needs to be in an aggregation expression context.

Since we're now supporting both expression contexts it's now possible to handle binary and unary comparisons on field names that contain `$` or `.` without erroring out. This PR does that by replacing uses of `safe_name` (which can throw errors) in `$match` stage building with `ColumnRef::from_comparison_target` (which does not throw errors).

Supporting aggregation expression contexts was also a blocker for column-to-column binary comparisons. So I added that support to this PR. But those comparisons don't support relationship paths for the time being.

[MDB-154](https://hasurahq.atlassian.net/browse/MDB-154)

[MDB-154]: https://hasurahq.atlassian.net/browse/MDB-154?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dmoverton (approved):
Looks good

PR #75, opened 2024-06-12
support root collection column references

hallettj:
Supports root collection column references correctly. We were using `$$ROOT` which just refers to the current collection. Root references actually reference the collection of the nearest query up the tree in the query request. The query plan that we produce from a query request is structured differently, and has `Query` values in positions where the request doesn't. So this PR introduces a "scope" concept in the query plan to capture the intended reference scopes from the request.

A couple of notes here on how the MongoDB expressions are evaluated:

The special variable `$$ROOT` refers to the "current document" being processed by the pipeline. When we look up a relation we use a `$lookup` stage which has it's own pipeline. If we reference `$$ROOT` _inside_ if the `$lookup` pipeline then it resolves to documents in the related collection - the one we are looking up.

On the other hand the `$lookup` stage also accepts a `let` map to bind variables. When we reference `$$ROOT` in a variable assignment in this `let` map it references the collection we are joining **from**, not the collection we are joining to. And then the variable bound in the `let` map is in scope in the `$lookup` pipeline.

[MDB-6](https://hasurahq.atlassian.net/browse/MDB-6)

[MDB-6]: https://hasurahq.atlassian.net/browse/MDB-6?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/crates/mongodb-agent-common/src/mongodb/selection.rs
+++ b/crates/mongodb-agent-common/src/mongodb/selection.rs
@@ -323,10 +323,10 @@ mod tests {
 
         let query_plan = plan_for_query_request(&students_config(), query_request)?;
 
-        // TODO: This selection illustrates that we end up looking up the relationship twice (once
-        // with the key `class_students`, and then with the key `class_students_0`). This is
-        // because the queries on the two relationships have different scope names. The query would
-        // work with just one lookup. Can we do that optimization?
+        // TODO: MDB-164 This selection illustrates that we end up looking up the relationship
+        // twice (once with the key `class_students`, and then with the key `class_students_0`).
+        // This is because the queries on the two relationships have different scope names. The
+        // query would work with just one lookup. Can we do that optimization?
         let selection = Selection::from_query_request(&query_plan)?;
         assert_eq!(
             Into::<Document>::into(selection),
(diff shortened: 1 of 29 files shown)


crates/mongodb-agent-common/src/mongodb/selection.rs
```diff
@@ -323,6 +323,10 @@ mod tests {
 
         let query_plan = plan_for_query_request(&students_config(), query_request)?;
 
+        // TODO: This selection illustrates that we end up looking up the relationship twice (once
+        // with the key `class_students`, and then with the key `class_students_0`). This is
+        // because the queries on the two relationships have different scope names. The query would
+        // work with just one lookup. Can we do that optimization?
```
codedmart, line 329: Can we create a ticket for this TODO as well please.

codedmart (approved):
👍

crates/mongodb-agent-common/src/mongodb/selection.rs
hallettj: Sure thing, I filed https://hasurahq.atlassian.net/browse/MDB-164

PR #79, opened 2024-06-13
Relax type requirements for primary uniqueness constraint

dmoverton:
Don't require _id field to have type ObjectId when generate primary uniqueness constraint

### Type
- [ ] highlight
- [ ] enhancement
- [x] bugfix
- [ ] behaviour-change
- [ ] performance-enhancement
- [ ] security-fix

### DIFF
--- a/crates/configuration/src/configuration.rs
+++ b/crates/configuration/src/configuration.rs
@@ -285,7 +285,11 @@ fn get_primary_key_uniqueness_constraint(
     // Check to make sure our collection's object type contains the _id field
     // If it doesn't (should never happen, all collections need an _id column), don't generate the constraint
     let object_type = object_types.get(collection_type)?;
-    let _id_field = object_type.fields.get("_id")?;
+    let id_field = object_type.fields.get("_id")?;
+    match &id_field.r#type {
+        schema::Type::Scalar(scalar_type) if scalar_type.is_comparable() => Some(()),
+        _ => None,
+    }?;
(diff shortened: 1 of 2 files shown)


crates/configuration/src/configuration.rs
```diff
@@ -282,14 +282,10 @@ fn get_primary_key_uniqueness_constraint(
...
-    // Check to make sure our collection's object type contains the _id objectid field
+    // Check to make sure our collection's object type contains the _id field
     // If it doesn't (should never happen, all collections need an _id column), don't generate the constraint
     let object_type = object_types.get(collection_type)?;
-    let id_field = object_type.fields.get("_id")?;
-    match &id_field.r#type {
-        schema::Type::Scalar(BsonScalarType::ObjectId) => Some(()),
-        _ => None,
-    }?;
+    let _id_field = object_type.fields.get("_id")?;
```
daniel-chambers, line 288: Shouldn't this be validated to at least be a scalar type (and shouldn't it only be scalar types with equality?)? 
What's the behaviour if the _id column is an object type? Does that break in v3-engine?

dmoverton: MongoDB itself doesn't care. The `_id` can have any type and all types are comparable. However, I don't want to potentially break the engine, so I've modified it to only allow scalar types for which we define an equality operator in the capabilities.

hallettj (approved):
I didn't that `_id` could be any type. But I guess if everything is schemaless it does make sense. Looks good!

PR #80, opened 2024-06-13
fix bson serialization test for case where double and int unify to double

hallettj:
In #77 I neglected to adjust our round-trip bson-to-json serialization proptest to handle cases where we treat ints as interchangeable with doubles. This change adds a custom equality test for use solely in tests that can compare ints and doubles.

### DIFF
--- a/crates/cli/src/introspection/type_unification.rs
+++ b/crates/cli/src/introspection/type_unification.rs
@@ -174,6 +174,16 @@ pub fn unify_object_types(
     merged_type_map.into_values().collect()
 }
 
+/// True iff we consider a to be a supertype of b.
+///
+/// Note that if you add more supertypes here then it is important to also update the custom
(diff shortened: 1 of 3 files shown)


crates/cli/src/introspection/type_unification.rs
```diff
@@ -174,6 +174,16 @@ pub fn unify_object_types(
     merged_type_map.into_values().collect()
 }
 
+/// True iff we consider a to be a supertype of b.
```
codedmart, line 177: Super minor spelling `iff`

codedmart (approved):

crates/cli/src/introspection/type_unification.rs
hallettj: "iff" is an abbreviation for "if and only if"

PR #82, opened 2024-06-19
Use field type in comparison target instead of column type

dmoverton:
## Describe your changes

Fixes a bug where comparison operator lookup for nested field comparisons was using the column type rather than the type of the nested field, resulting in an `UnknownComparisonOperator` error.

Note: we will need to do the same thing for ordering and aggregates eventually.

[MDB-53](https://hasurahq.atlassian.net/browse/MDB-53)

### Type
- [ ] highlight
- [ ] enhancement
- [x] bugfix
- [ ] behaviour-change
- [ ] performance-enhancement
- [ ] security-fix

[MDB-53]: https://hasurahq.atlassian.net/browse/MDB-53?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ This changelog documents the changes between release versions.
 
 ## [Unreleased]
 
+- Fix bug with operator lookup when filtering on nested fields ([#82](https://github.com/hasura/ndc-mongodb/pull/82))
+
 ## [0.1.0] - 2024-06-13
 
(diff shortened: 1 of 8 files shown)


hallettj (approved):
Very nice!

PR #84, opened 2024-07-02
add config option for relaxed or canonical extended json output

hallettj:
Adds an option to allow users to opt into "relaxed" mode for Extended JSON output. Keeps "canonical" mode as the default because it is lossless. For example relaxed mode does not preserve the exact numeric type of each numeric value, while canonical mode does.

This does not affect inputs. For example sorts and filters will accept either canonical or relaxed input modes as before.

[MDB-169](https://hasurahq.atlassian.net/browse/MDB-169)

[MDB-169]: https://hasurahq.atlassian.net/browse/MDB-169?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


codedmart (approved):
Did we go with relaxed be default?

hallettj:
> Did we go with relaxed be default?

Per discussion from the meeting, we're sticking with strict by default to prioritize avoiding any potential data loss

PR #86, opened 2024-07-09
Update to ndc-spec v0.1.5 and ndc-sdk-rs v0.2.1

dmoverton:
## Describe your changes

Update to latest ndc-spec and sdk

### DIFF
--- a/crates/configuration/src/schema/mod.rs
+++ b/crates/configuration/src/schema/mod.rs
@@ -12,7 +12,7 @@ use crate::{WithName, WithNameRef};
 pub struct Collection {
     /// The name of a type declared in `objectTypes` that describes the fields of this collection.
     /// The type name may be the same as the collection name.
-    pub r#type: String,
+    pub r#type: ndc_models::ObjectTypeName,
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub description: Option<String>,
 }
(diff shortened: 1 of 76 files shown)


crates/configuration/src/schema/mod.rs
```diff
@@ -69,13 +69,15 @@ impl From<Type> for ndc_models::Type {
                 // ExtendedJSON can respresent any BSON value, including null, so it is always nullable
                 Type::ExtendedJSON => ndc_models::Type::Nullable {
                     underlying_type: Box::new(ndc_models::Type::Named {
-                        name: mongodb_support::EXTENDED_JSON_TYPE_NAME.to_owned(),
+                        name: mongodb_support::EXTENDED_JSON_TYPE_NAME.to_owned().into(),
```
hallettj, line 72: I think for cases like this we might avoid heap allocation with something like,

```rust
TypeName(SmolStr::new_static(mongodb_support::EXTENDED_JSON_TYPE_NAME))
```

That would be a bit easier if the ndc-models `newtype` macro added implementations for `From SmolStr` or `From &'static str`. But it looks like the ref-cast crate helps to fill that gap?

This is absolutely not worth you spending time on making changes. I'm commenting only because I enjoy the allocation golf thought exercises.

crates/mongodb-agent-common/src/aggregation_function.rs
```diff
@@ -28,7 +28,7 @@ impl AggregationFunction {
         all::<AggregationFunction>()
             .find(|variant| variant.graphql_name() == s)
             .ok_or(QueryPlanError::UnknownAggregateFunction {
-                aggregate_function: s.to_owned(),
+                aggregate_function: s.to_owned().into(),
```
hallettj, line 31: It looks like we should at least add a `From &'a str` to `newtype`

hallettj (approved):
This is very helpful, thanks for all this work!

PR #88, opened 2024-07-10
update integration fixtures to work with latest cli

hallettj:
I reorganized and updated the ddn and connector configuration fixtures used in integration tests to match the structure used by the ddn cli. This lets us use commands like `ddn connector introspect` to update connector configuration, and `ddn connector-link update` to update subgraphs.

The fixtures are tied to the arion services, so those commands will work after running services with `arion up -d`.

I added `fixtures/hasura/README.md` with the exact commands to run for updates.


fixtures/hasura/chinook/metadata/models/Customer.hml
```diff
@@ -7,31 +7,31 @@ definition:
     - name: id
       type: Chinook_ObjectId!
     - name: address
-      type: String!
```
hallettj, line 10: Lots of types have been changed from non-nullable to nullable which I think should not have happened. I want to investigate that.

dmoverton: We changed the default, didn't we?

hallettj: I have `allSchemaNullable: false` so I'd expect non-nullable types. But I did file a ticket because introspection on the sample_mflix connector changes the setting to `true` instead of leaving it as-is. That hasn't happened for the chinook connector configuration. I don't know what the difference is. https://hasurahq.atlassian.net/browse/MDB-172

hallettj: Never mind about that bug - it was user error.

To get the configuration I wanted to stick I needed to put a newer modification time on the configuration file, and I needed to make it valid JSON. The things I tried were to revert the change using git, which I think rolled back the modification time, and when I edited by hand I put in a trailing comma which was not valid JSON.

hallettj: Never mind, it looks like these nullable changes are also expected behavior. I was too tired when I was looking over this yesterday evening

dmoverton (approved):
Looks good to me.

PR #92, opened 2024-07-19
accept predicate arguments

hallettj:
Accept predicate arguments in native mutations and native queries.

Unfortunately this needs more work - the code for parsing predicate values is incorrect. I expected to receive a value of type `ndc_models::Expression`, but in the integration test in this PR the predicate value is:

```json
{"albumId":{"_eq":3}}
```

The parsing code is in the `plan_for_predicate` function in `crates/ndc-query-plan/src/plan_for_query_request/plan_for_arguments.rs`. ~~Somehow that value needs to be turned into an `Expression`. Unless there is a parser already written for this predicate format we'll have to make one.~~

**Edit:** I'm told the predicate value is supposed to be a serialized `Expression`. This might be a bug in the engine. If so this PR is ready to go if that bug can be resolved, and the tests pass.

I moved logic that had previously been implemented in the connector into ndc-query-plan, creating `plan_for_mutation_request` in the process. Parsing predicates, and matching up types to arguments is now done in database-agnostic code in ndc-query-plan.

So `ndc_models::Type` has a `Predicate` variant, but I chose not to add a predicate variant to `ndc_query_plan::Type`. That is because if I did there would be a number of cases where I would have to error out in cases where we have values, but where a predicate doesn't make sense. I don't think predicates are query-time values anyway - they only apply at query **build** time. They aren't stored in databases, they can't be given in variables, they aren't returned in responses.

Following the philosophy of making invalid states unrepresentable I kept the `ndc_query_plan::Type` as-is, and added variants to the `Argument` types instead to distinguish predicates from query-time values. For example here is the new version of `ndc_query_plan::Argument`:

```rs
pub enum Argument<T: ConnectorTypes> {
    /// The argument is provided by reference to a variable
    Variable {
        name: ndc::VariableName,
        argument_type: Type<T::ScalarType>,
    },
    /// The argument is provided as a literal value
    Literal {
        value: serde_json::Value,
        argument_type: Type<T::ScalarType>,
    },
    /// The argument was a literal value that has been parsed as an [Expression]
    Predicate { expression: Expression<T> },
}
```

There are similar changes to `RelationArgument`, and to a new type, `MutationProcedureArgument`.

Completes https://linear.app/hasura/issue/NDC-175/accept-predicate-as-an-argument-type-with-native-mutations


crates/ndc-query-plan/src/plan_for_query_request/query_plan_state.rs
```diff
@@ -159,18 +170,7 @@ impl<T: QueryContext> QueryPlanState<'_, T> {
...
-    pub fn register_variable_use_of_unknown_type(&mut self, variable_name: &ndc::VariableName) {
-        self.register_variable_use_helper(variable_name, None)
-    }
-
-    fn register_variable_use_helper(
-        &mut self,
-        variable_name: &ndc::VariableName,
-        expected_type: Option<Type<T::ScalarType>>,
-    ) {
+        // self.register_variable_use_helper(variable_name, Some(expected_type))
```
dmoverton, line 173: Why is this commented out?

hallettj: Oh good catch! There was a point where I had split out a helper function to cover some of that function's logic, but then I ended up keeping everything in one function.

dmoverton (approved):
Looks good

PR #93, opened 2024-08-02
switch sample_mflix fixture config to use new boolean expression metadata

hallettj:
These changes only affect integration tests and the preconfigured dev environment. The fixture metadata configures two MongoDB connectors: one using MongoDB's sample_mlfix sample data set, the other using a Chinook data set.

Updates object types and scalar types in metadata configuration for both connectors to use `BooleanExpressionType` instead of `ObjectBooleanExpressionType`.

I also moved the scalar type metadata so that it is shared between the sample_mflix and chinook connectors. Now we are using the same set of scalar types with both connectors instead of having `Chinook_` -prefixed types for the chinook connector.

References [NDC-381](https://linear.app/hasura/issue/NDC-381/update-metadata-fixtures-to-configure-aggregations)


dmoverton (approved):
Looks good!

PR #95, opened 2024-08-08
add aggregate expression types to fixture metadata

hallettj:
Once again, this only affects integration tests and the local dev environment.

Adds aggregation expression types for everything that looks it should be aggregatable. The connector isn't advertising support for nested field aggregates yet so the nested field aggregates are commented out for now. I filed [NDC-386](https://linear.app/hasura/issue/NDC-386/update-capabilities-to-support-nested-aggregates) to follow up on that. Everything looks good in ad-hoc testing. I'm planning to follow up with integration tests shortly.

Completes [NDC-381](https://linear.app/hasura/issue/NDC-381/update-metadata-fixtures-to-configure-aggregations)

### DIFF
--- a/fixtures/hasura/common/metadata/scalar-types/Int.hml
+++ b/fixtures/hasura/common/metadata/scalar-types/Int.hml
@@ -60,3 +60,40 @@ definition:
     enable: true
   graphql:
     typeName: IntComparisonExp
+
+---
+kind: AggregateExpression
+version: v1
+definition:
+  name: IntAggregateExp
+  operand:
+    scalar:
+      aggregatedType: Int
+      aggregationFunctions:
+        - name: _avg
+          returnType: Int
+        - name: _max
+          returnType: Int
+        - name: _min
+          returnType: Int
+        - name: _sum
+          returnType: Int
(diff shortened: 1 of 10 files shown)


fixtures/hasura/common/metadata/scalar-types/Int.hml
```diff
@@ -60,3 +60,40 @@ definition:
...
+kind: AggregateExpression
+version: v1
+definition:
+  name: IntAggregateExp
+  operand:
+    scalar:
+      aggregatedType: Int
+      aggregationFunctions:
+        - name: _avg
+          returnType: Int
```
hallettj, line 74: The return type for averages over `Int` should be `Float`. We need to fix that in the connector capabilities. I filed an issue to follow up on this, https://linear.app/hasura/issue/NDC-385/return-type-for-average-of-ints-should-be-float-not-int

daniel-chambers (approved):
Seems reasonable to me!

You should also consider adding aggregates to your relationships as well, so you can get `Customer.Invoices_aggregate`-style fields.

dmoverton (approved):

PR #96, opened 2024-08-13
serialize aggregate results as simple instead of extended json

hallettj:
This makes serialization for aggregate results behave the same as serialization for non-aggregate query responses.


dmoverton (approved):
Looks good

PR #99, opened 2024-08-22
support all comparison and aggregation functions on extended json fields

hallettj:
Since Extended JSON fields may contain any data users may reasonably want access to all comparison and aggregation operations. MongoDB is designed to accommodate functions that don't make sense on all values they are applied to. I included a native query that produces a collection of mixed data types, and some integration tests on that data.

I also updated the dev shell to pull the cli from the [nix flake](https://github.com/hasura/ddn-cli-nix) that @TheInnerLight helpfully set up.

### DIFF
--- a/crates/integration-tests/src/tests/aggregation.rs
+++ b/crates/integration-tests/src/tests/aggregation.rs
@@ -43,31 +43,59 @@ async fn aggregates_extended_json_representing_mixture_of_numeric_types() -> any
     assert_yaml_snapshot!(
         graphql_query(
             r#"
-                query {
-
-                  track(order_by: { id: Asc }, where: { albumId: { _eq: $albumId } }) {
-                    milliseconds
-                    unitPrice
-                  }
-                  trackAggregate(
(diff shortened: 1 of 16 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,14 @@ This changelog documents the changes between release versions.
 
 ## [Unreleased]
 
+### Added
+
+- Extended JSON fields no support all comparison and aggregation functions ([#99](https://github.com/hasura/ndc-mongodb/pull/99))
```
dmoverton, line 9:
```suggestion
- Extended JSON fields now support all comparison and aggregation functions ([#99](https://github.com/hasura/ndc-mongodb/pull/99))
```

dmoverton (approved):
Looks good, apart from the typo in the changelog. We should also allow `order_by` on extended json fields, but that could be in a separate PR.

hallettj:
> Looks good, apart from the typo in the changelog. We should also allow `order_by` on extended json fields, but that could be in a separate PR.

Good point @dmoverton. This PR does add order_by functions, but I didn't put in a test for that. I just added one.

PR #100, opened 2024-08-22
ci: update rust-sec advisory db in every test run

hallettj:
We have had a security vulnerability check on cargo dependencies in place for a long time. But it's necessary to update the advisory db to get the latest advisories. This change updates a github workflow to run the update on every CI test run.

I tested that the check catches problems by running a test with `gix-fs` installed at version 0.10.2. See https://rustsec.org/advisories/RUSTSEC-2024-0350.html

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -30,6 +30,9 @@ jobs:
       - name: run linter checks with clippy 🔨
         run: nix build .#checks.x86_64-linux.lint --print-build-logs
 
+      - name: update rust-sec advisory db before scanning for vulnerabilities
+        run: nix flake lock --update-input advisory-db
+
       - name: audit for reported security problems 🔨
         run: nix build .#checks.x86_64-linux.audit --print-build-logs
 


codedmart:
Is this the correct command and output? Why are we ignoring `yanked` and why is it showing errors, but the action passes.
image

dmoverton (approved):

hallettj:
> Is this the correct command and output? Why are we ignoring `yanked` and why is it showing errors, but the action passes.

Yes, that's correct. That's the way [crane](https://crane.dev/) has it set up. Since we're building with crane we get that audit configuration for free.

Failed "yanked" checks are a limitation of the Nix setup. The check runs in a sandbox without network access, and the downloaded advisory database doesn't include "yanked" status. But we do get advisories on vulnerabilities whether or not a crate version was yanked, so the audit check works as intended.

Running the audit in a Nix sandbox has the advantage that we cache successful checks. If the hashes of `Cargo.lock` and the advisory database are unchanged then CI uses the cached success result. We could move the check out of the Nix sandbox, and do live network requests instead. But it would be slower, and I don't think we would gain anything.

I don't think the `--ignore yanked` option actually does anything. `--ignore` is for ignoring specific advisory IDs, like `--ignore RUSTSEC-2017-0001`. The CI check gives the same results with or without that option. My guess is the crane dev was hoping to suppress those error messages, but it didn't work.

codedmart (approved):

PR #105, opened 2024-09-24
cli: add native-query create subcommand to create native query configuration

hallettj:
This is work in progress. The new subcommand is hidden by a build-time feature flag, `native-query-subcommand`, so that the work that is done so far can be reviewed and merged.

The new command accepts a name for a native query, an optional input collection, and a file containing an aggregation pipeline, and produces a native query configuration file that includes type declarations inferred from the pipeline.

Here is an example that uses the sample-mflix connector configuration in the repo. Create a file called `pipeline.json` containing this pipeline:

```json
[
  {
    "$replaceWith": {
      "title_words": { "$split": ["$title", " "] }
    }
  },
  { "$unwind": { "path": "$title_words" } },
  {
    "$group": {
      "_id": "$title_words",
      "count": { "$count": {} }
    }
  }
]
```

Then run this command:

```sh
$ cargo run -p mongodb-cli-plugin --features native-query-subcommand -- -\
  -p fixtures/hasura/sample_mflix/connector/ \
  native-query create pipeline.json --name title_words --collection movies
```

That should create `fixtures/hasura/sample_mflix/connector/native_queries/title_words.json`

There are notes on what is and isn't done in https://linear.app/hasura/project/mongodb-automatic-native-queries-fceb7aed3196/overview. Here is a copy of those notes:

These CLI commands will let users create native queries without having to write type declarations themselves. The CLI infers the parameter list, and return document type. We will want a type annotation syntax for arguments, but it may be possible to infer parameter types in some cases.

Here is what's done:

- CLI command to generate native query configuration without support for parameters, supports a subset of aggregation pipeline features
- supported pipeline stages:
  - `$documents`
  - `$match`
  - `$sort`
  - `$limit`
  - `$skip`
  - `$group`
  - `$replaceWith`
  - `$unwind`
- string syntax for references to input document fields is supported, e.g. `"$tomatoes.viewer.rating"`
- supported aggregation operators:
  - `$split`


crates/cli/src/native_query/aggregation_expression.rs
```diff
@@ -0,0 +1,131 @@
...
+        Bson::Timestamp(_) => todo!(),
+        Bson::Binary(_) => todo!(),
+        Bson::ObjectId(_) => todo!(),
+        Bson::DateTime(_) => todo!(),
+        Bson::Symbol(_) => todo!(),
+        Bson::Decimal128(_) => todo!(),
+        Bson::Undefined => todo!(),
+        Bson::MaxKey => todo!(),
+        Bson::MinKey => todo!(),
+        Bson::DbPointer(_) => todo!(),
```
codedmart, line 43: Curious why the other number types are still `todo!` other then `Double`?  Aren't the other number types pretty easy to add here?

codedmart (approved):
:+1:

crates/cli/src/native_query/aggregation_expression.rs
hallettj: Yes, it very easy to add implementations for the other scalar types. It just hasn't made it to the top of my todo list.

PR #107, opened 2024-09-24
accept queries that make array-to-scalar comparisons

hallettj:
**Important**: this change allows filtering documents by scalar values inside arrays. But it does so using behavior of the GraphQL Engine that is currently undefined. We are in the process of defining that behavior - in a future Engine update to keep array-of-scalar comparisons working it will be necessary to configure additional metadata, and it will be necessary to upgrade the MongoDB connector at the same time. We expect that GraphQL APIs will remain the same before and after that Engine update.

MongoDB allows comparing array values to scalars. For example,

```ts
db.movies.aggregate([{ $match: { cast: { $eq: "Albert Austin" } } }])
```

The condition passes if any array element passes the comparison.

Currently the GraphQL Engine does not have a way to filter by scalar values in array fields. But it does allow declaring that array fields can be used in scalar comparisons. So with some connector updates this allows us to unblock users who need to be able to filter by array values.

This PR updates the connector to:
- find a scalar comparison operator definition for equality corresponding to the scalar type of array elements in the given column
- infer the appropriate scalar type for the comparison value instead of assuming that the value type in an equality comparison is the same as the column type

Some notes on the implementation and implications:

The connector needs to know operand types for comparisons. The left operand is always a database
document field so we can get that type from the schema. But if the right operand is an inline value
or a variable we have to infer its type from context.

Normally we assume that the right operand of Equal is the same type as the left operand. But since
MongoDB allows comparing arrays to scalar values then in case that is desired then instead of
inferring an array type for the right operand to match the left, we want the type of the right
operand to be the array element type of the left operand. That brings up the question: how do we
know if the user's intention is to make an array-to-scalar comparison, or an array-to-array
comparison? Since we don't support array-to-array comparisons yet for simplicity this PR assumes
that if the field has an array type, the right-operand type is a scalar type.

When we do support array-to-array comparisons we have two options.

1. inspect inline values or variable values given for the right operand, and assume an
  array-to-scalar comparison only if all of those are non-array values
2. or get the GraphQL Engine to include a type with `ComparisonValue` in which case we can
  use that as the right-operand type

It is important that queries behave the same when given an inline value or variables. So we can't
just check inline values, and punt on variables. It will require a little more work to thread
variables through the code to the point they are needed for this check so I haven't done that just
yet.

**Edit:** We have plans to update the NDC spec to get the necessary type information from the Engine in a future version.

### DIFF
--- a/crates/integration-tests/src/tests/filtering.rs
+++ b/crates/integration-tests/src/tests/filtering.rs
@@ -1,6 +1,7 @@
 use insta::assert_yaml_snapshot;
+use ndc_test_helpers::{binop, field, query, query_request, target, variable};
 
-use crate::graphql_query;
+use crate::{connector::Connector, graphql_query, run_connector_query};
 
 #[tokio::test]
 async fn filters_on_extended_json_using_string_comparison() -> anyhow::Result<()> {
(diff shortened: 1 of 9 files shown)


hallettj:
I included a test that uses query request variables, which is failing. It comes down to an inconsistency in MongoDB's behavior:

```
sample_mflix> db.movies.aggregate([{ $match: { cast: { $eq: "Albert Austin" } } }, { $count: "rows" }])
[ { rows: 1 } ]

sample_mflix> db.movies.aggregate([{ $match: { $expr: { $eq: ["$cast", "Albert Austin"] } } }, { $count: "rows" }])
< no rows >
```

We need to use the second syntax when comparing to variables because variables cannot be referenced in a query document (like the argument to `$match` in the first query). But because these two forms behave differently existential comparison over arrays won't work in remote joins.

crates/integration-tests/src/tests/filtering.rs
```diff
@@ -76,6 +76,17 @@ async fn filters_by_comparisons_on_elements_of_array_of_scalars() -> anyhow::Res
...
+    //
+    // > Kind: Command failed: Error code 5491300 (Location5491300): $documents' is not allowed in user requests, labels: {}
+    //
+    // This doesn't affect native queries that don't use the $documents stage.
+    if let Ok(image) = std::env::var("MONGODB_IMAGE") {
+        if image == "mongo:5" {
+            return Ok(());
+        }
+    }
+
```
codedmart, line 89: Do we still need mongodb 5 tests? We stopped supporting it. In the readme we state 6 and above. Also 5 is EOL Oct 2024.

hallettj: I think it would make sense to disable v5 testing in another PR

codedmart (approved):

PR #108, opened 2024-09-27
fix selecting nested field with name that begins with dollar sign

hallettj:
This PR updates the logic for building selection documents for the `$replaceWith` pipeline stage to be more rigorous. It uses an expanded `ColumnRef` enum to keep track of whether the selected reference has a name that needs to be escaped, is a variable, etc.

### DIFF
--- a/crates/mongodb-agent-common/src/mongodb/selection.rs
+++ b/crates/mongodb-agent-common/src/mongodb/selection.rs
@@ -1,11 +1,13 @@
 use indexmap::IndexMap;
 use mongodb::bson::{self, doc, Bson, Document};
+use ndc_models::FieldName;
 use serde::{Deserialize, Serialize};
 
 use crate::{
     interface_types::MongoAgentError,
     mongo_query_plan::{Field, NestedArray, NestedField, NestedObject, QueryPlan},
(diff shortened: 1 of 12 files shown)


crates/mongodb-agent-common/src/mongodb/selection.rs
```diff
@@ -161,33 +158,43 @@ fn selection_for_field(parent_columns: &[&str], field: &Field) -> Result<Bson, M
...
+    }
 }
 
+// fn append_to_path<'a, 'b, 'c, T>(parent_columns: &'a [&'b T], column: &'c T) -> Vec<&'c T>
+// where
+//     'b: 'c,
+// {
+//     parent_columns.iter().copied().chain(Some(column)).collect()
+// }
+
```
codedmart, line 197: Did you leave this intentionally or can it be cleaned up?

codedmart (approved):

crates/mongodb-agent-common/src/mongodb/selection.rs
hallettj: Thanks for catching that - I removed it

PR #113, opened 2024-10-08
escape field names in relation mappings when necessary

hallettj:
Previously on attempts to join on field names that require escaping the request would fail with an error saying that a name was not valid. This change does the necessary escaping to make the join work.

This is the last change that was needed to get rid of `safe_name` which was the quick-and-easy fail-on-problematic-names solution that we used early on.


crates/mongodb-agent-common/src/query/relations.rs
```diff
@@ -804,114 +896,125 @@ mod tests {
...
-    //         }]),
-    //     );
-    //
-    //     let result = execute_query_request(db, &mflix_config(), query_request).await?;
-    //     assert_eq!(expected_response, result);
-    //
-    //     Ok(())
-    // }
+    #[tokio::test]
+    async fn filters_by_field_nested_in_object_in_related_collection() -> Result<(), anyhow::Error>
```
hallettj, line 900: I found this ancient, forgotten test so I uncommented it.

crates/mongodb-agent-common/src/query/relations.rs
```diff
@@ -66,38 +66,60 @@ fn make_lookup_stage(
...
     scope: Option<&Scope>,
-) -> Result<Stage> {
-    // If we are mapping a single field in the source collection to a single field in the target
-    // collection then we can use the correlated subquery syntax.
-    if column_mapping.len() == 1 {
+) -> Stage {
+    // If there is a single column mapping, and the source and target field references can be
+    // expressed as match keys (we don't need to escape field names), then we can use a concise
+    // correlated subquery. Otherwise we need to fall back to an uncorrelated subquery.
+    let safe_single_column_mapping = if column_mapping.len() == 1 {
```
hallettj, line 73: The meat of the changes starts here

crates/integration-tests/src/lib.rs
```diff
@@ -35,6 +36,12 @@ fn get_connector_chinook_url() -> anyhow::Result<Url> {
     Ok(url)
 }
 
+fn get_connector_test_cases_url() -> anyhow::Result<Url> {
+    let input = env::var(CONNECTOR_TEST_CASES_URL).map_err(|_| anyhow!("please set {CONNECTOR_TEST_CASES_URL} to the the base URL of a running MongoDB connector instance"))?;
```
daniel-chambers, line 40:
```suggestion
    let input = env::var(CONNECTOR_TEST_CASES_URL).map_err(|_| anyhow!("please set {CONNECTOR_TEST_CASES_URL} to the base URL of a running MongoDB connector instance"))?;
```

daniel-chambers (approved):
LGTM!

PR #116, opened 2024-10-23
infer types for parameters based on usage when generating native query

hallettj:
This gets the functionality in place - this change demonstrates detecting parameters and inferring types for parameters used in equality comparisons in a  `$match` stage. It also detects parameters in aggregation expressions in any context, but does not infer types for those yet. We'll flesh type inference in more contexts as we flesh out support for more aggregation stages, operators, etc.

All of this is still behind a feature flag, `native-query-subcommand`. This work is still in development and isn't ready for stable use yet.


crates/cli/src/native_query/mod.rs
```diff
@@ -284,6 +305,34 @@ mod tests {
...
+            unordered_elements_are![(
+                displays_as(eq("title")),
+                field!(
+                    ObjectField.r#type,
+                    eq(&Type::Scalar(BsonScalarType::String))
+                )
+            )]
+        );
+        Ok(())
+    }
```
hallettj, line 334: This test demonstrates the magic

codedmart (approved):

PR #118, opened 2024-10-29
prune superfluous inferred object types from native query configuration

hallettj:


crates/cli/src/native_query/helpers.rs
```diff
@@ -24,3 +27,32 @@ pub fn find_collection_object_type(
...
+    let re = Regex::new(r"^(.*?)_(\d+)$").unwrap();
+    let Some(captures) = re.captures(name) else {
+        return (Cow::Borrowed(name), 0);
+    };
+    let prefix = captures.get(1).unwrap().as_str();
+    let Some(count) = captures.get(2).and_then(|s| s.as_str().parse().ok()) else {
+        return (Cow::Borrowed(name), 0);
+    };
+    (Cow::Owned(prefix.to_string()), count)
+}
```
codedmart, line 58: So a suffix only gets added if there is more then one? Or does this not really matter because the prune fns below remove these?

codedmart (approved):

crates/cli/src/native_query/helpers.rs
hallettj: Yes, the suffix is only added if the count is 1 or higher. And then the prune step attempts to remove the suffix if that wouldn't cause a collision, but otherwise uses the lowest-number suffix that doesn't cause a collision after pruning.

PR #120, opened 2024-11-07
support more aggregation operators when generating native query configurations

hallettj:
Adds or refines support for these operators:

Arithmetic Expression Operators

- `$abs`
- `$add`
- `$divide`
- `$multiply`
- `$subtract`

Array Expression Operators

- `$arrayElemAt`

Boolean Expression Operators

- `$and`
- `$not`
- `$or`

Comparison Expression Operators

- `$eq`
- `$gt`
- `$gte`
- `$lt`
- `$lte`
- `$ne`

Set Expression Operators

- `$allElementsTrue`
- `$anyElementTrue`

String Expression Operators

- `$split`

Trigonometry Expression Operators

- `$sin`
- `$cos`
- `$tan`
- `$asin`
- `$acos`
- `$atan`
- `$asinh`
- `$acosh`
- `$atanh`
- `$sinh`
- `$cosh`
- `$tanh`

Accumulators (`$group`, `$bucket`, `$bucketAuto`, `$setWindowFields`)

- `$avg`
- `$count`
- `$max`
- `$min`
- `$push`
- `$sum`

Also improves type inference to make all of these operators work.

This is work an an in-progress feature that is gated behind a feature flag, `native-query-subcommand`

### DIFF
--- a/crates/cli/src/native_query/type_constraint.rs
+++ b/crates/cli/src/native_query/type_constraint.rs
@@ -53,8 +53,19 @@ pub enum TypeConstraint {
     },
 
     // Complex types
+    
     Union(BTreeSet<TypeConstraint>),
-    OneOf(BTreeSet<TypeConstraint>), // unlike union type should be only one of these, but we don't know which one
+
+    /// Unlike Union we expect the solved concrete type for a variable with a OneOf constraint may
+    /// be one of the types in the set, but we don't know yet which one. This is useful for MongoDB
+    /// operators that expect an input of any numeric type. We use OneOf because we don't know
+    /// which numeric type to infer until we see more usage evidence of the same type variable.
+    ///
+    /// In other words with Union we have specific evidence that a variable occurs in contexts of
+    /// multiple concrete types, while with OneOf we **don't** have specific evidence that the
+    /// variable takes multiple types, but there are multiple possibilities of the type or types
+    /// that it does take.
+    OneOf(BTreeSet<TypeConstraint>),
 
     /// Indicates a type that is the same as the type of the given variable.
     Variable(TypeVariable),


crates/cli/src/native_query/type_constraint.rs
```diff
@@ -16,28 +16,46 @@ impl TypeVariable {
...
     Object(ObjectTypeName),
     ArrayOf(Box<TypeConstraint>),
-    Nullable(Box<TypeConstraint>),
     Predicate {
         object_type_name: ObjectTypeName,
     },
 
+    // Complex types
+    Union(BTreeSet<TypeConstraint>),
+    OneOf(BTreeSet<TypeConstraint>), // unlike union type should be only one of these, but we don't know which one
```
danieljharvey, line 57: Could you provide a concrete example here in a comment perhaps? This may be entirely a reflection of my lack of understanding around Mongo's semantics, but I can't get my head around when this would be used.

crates/cli/src/native_query/type_solver/mod.rs
```diff
@@ -24,73 +22,72 @@ pub fn unify(
...
 
-    // TODO: This could be simplified. Instead of mutating constraints using `simplify_constraints`
-    // we might be able to roll all constraints into one and pass that to `constraint_to_type` in
-    // one step, but leave the original constraints unchanged if any part of that fails. That could
-    // make it simpler to keep track of source locations for when we want to report type mismatch
-    // errors between constraints.
     loop {
         let prev_type_variables = type_variables.clone();
         let prev_solutions = solutions.clone();
+        let prev_substitutions = substitutions.clone();
```
danieljharvey, line 43: I'm glad your simplification was instead to make a clone of the substitutions for backtracking.

crates/cli/src/native_query/aggregation_expression.rs
```diff
@@ -74,21 +127,185 @@ fn infer_type_from_aggregation_expression_document(
...
+    }
+}
+
+fn homogeneous_binary_operator_operand_type(
+    context: &mut PipelineTypeContext<'_>,
+    desired_object_type_name: &str,
+    operand_type_hint: Option<TypeConstraint>,
+    operator: &str,
+    operand: Bson,
+) -> Result<TypeConstraint> {
```
danieljharvey, line 278: So, broadly, forgive my massive handwaving, do I understand correctly that `TypeConstraint` contains either a concrete type or information we have about an unknown type (ie, that it's `Numeric`)?

And that we use subtyping to move constraints towards types when we combine them?

crates/cli/src/native_query/aggregation-operator-progress.md
```diff
@@ -0,0 +1,280 @@
+Arithmetic Expression Operators
+
+- [x] $abs - Returns the absolute value of a number.
```
danieljharvey, line 3: I assume that Mongo has no deep concept of introspection for these functions, so it's just a case of special casing each one we want to support?

danieljharvey (approved):
I've left some questions to clarify my understanding, however I do not think they should stop us merging this work.

crates/cli/src/native_query/type_solver/simplify.rs
```diff
@@ -285,80 +273,74 @@ fn solve_scalar(
...
-    }
-
-    if a == b || is_supertype(&a, &b) {
-        Ok(C::Scalar(a))
-    } else if is_supertype(&b, &a) {
-        Ok(C::Scalar(b))
-    } else {
-        Err((C::Scalar(a), C::Scalar(b)))
+    match variance {
+        Variance::Covariant => {
```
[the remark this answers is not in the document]
hallettj: Oh you're right - I want to return the supertype in the covariant case, and the subtype in the contravariant case. But that's not actually how I wrote it is it?

crates/cli/src/native_query/type_constraint.rs
hallettj: That's a good idea, I'll add more details in a comment.

OneOf is helpful mostly for numeric inputs, where we're expecting a numeric type, but we don't know specifically which one. We probably want to infer a specific numeric type instead of a union of all numeric types, which would present as `ExtendedJSON` (the top type) in the API which isn't as helpful.

OTOH my thinking is Union is for cases where we have specific evidence a type variable is associated with multiple distinct concrete types. I'm mainly using Union to indicate nullability.

Another way to handle OneOf that could be less handwavy could be to use `SubtypeOf(Union(...))`. But I'm not sure exactly how the semantics of that would play out.

crates/cli/src/native_query/type_solver/constraint_to_type.rs
```diff
@@ -88,6 +81,37 @@ pub fn constraint_to_type(
...
+                added_object_types,
+                object_type_constraints,
+                underlying_constraint,
+            )?
+            .map(|t| Type::Nullable(Box::new(t)))
+        }
+
+        C::Union(_) => Some(Type::ExtendedJSON),
+
+        t @ C::OneOf(_) if t.is_numeric() => {
```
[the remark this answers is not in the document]
hallettj: Yes, good point! Type annotations will show up as another constraint in the set of constraints for the type variable. So before we get to `constraint_to_type` we'll see a call to `simplify_pair` with `OneOf(...)` and `Scalar(_)` (or whatever concrete type is given by the annotation) which will simplify to the annotation type if that overlaps with one of the types in the `OneOf` set. If there is no overlap we'll either report an error, or have some way of signalling that annotations suppress type mismatches.

crates/cli/src/native_query/aggregation_expression.rs
hallettj: Yes, that's pretty much it with the addendum that a `TypeConstraint` may reference a type variable. We have a type variable for the query result type, and another type variable for each parameter, and each variable is associated with a set of `TypeConstraint`s. Since `TypeConstraint` can reference variables that's a way for variables to indirectly reference each other.

Subtyping is part of it. I'm considering the query result variable to be covariant so multiple constraints converge on the supertype. Parameters are contravariant so those constraints should converge on the subtype, or lead to a mismatch error. But constraints can also relate variables in ways that are not related to subtyping. For example we've got `FieldOf` which is a way to state that the type of a variable is the same as the type of a given field of a given object type.

crates/cli/src/native_query/aggregation-operator-progress.md
hallettj: Unfortunately MongoDB doesn't have much of a concept of typing at all. I don't think there is any way to introspect for signatures for functions like these.

It's possible we could build our own function signature language, and write signatures ourselves. But I thought it would be easier to hard-code this stuff, especially since the query language has a closed, limited set of functions.

hallettj:
> I've left some questions to clarify my understanding, however I do not think they should stop us merging this work.

Thanks very much for the review and the feedback!

crates/cli/src/native_query/aggregation-operator-progress.md
danieljharvey: I think this approach is absolutely fine, it's very clear what is going on which is great.

PR #121, opened 2024-11-11
support more query operators in native query pipeline type inference

hallettj:
This is work an an in-progress feature that is gated behind a feature flag, `native-query-subcommand`

When generating configurations for native queries it is necessary to analyze operators in `$match` stages in aggregation pipelines to infer types for any parameters that appear in operator arguments.

This PR adds the logic to process these operators:

- `$and` | `$or` | `$nor`
- `$not`
- `$elemMatch`
- `$eq` | `$ne` | `$gt` | `$lt` | `$gte` | `$lte`
- `$in` | `$nin`
- `$exists`
- `$type`
- `$mod`
- `$regex`
- `$all`
- `$size`

The full set of available operators is given here: https://www.mongodb.com/docs/manual/reference/operator/query/


crates/cli/src/native_query/pipeline/match_stage.rs
```diff
@@ -41,61 +41,185 @@ fn check_match_doc_for_parameters_helper(
...
     field_type: &TypeConstraint,
-    match_doc: Document,
+    operator: String,
+    match_expression: Bson,
 ) -> Result<()> {
-    for (operator, match_expression) in match_doc {
-        match operator.as_ref() {
-            "$eq" => analyze_match_expression(
+    match operator.as_ref() {
+        "$and" | "$or" | "$nor" => {
```
danieljharvey, line 94: So once again we hardcode every function we want to support? This does at least result in very concrete checking code.

danieljharvey (approved):
Makes sense.

PR #122, opened 2024-11-14
add $in and $nin operators to connector schema

hallettj:
Adds `$in` and `$nin` to all scalar types that also have `$eq` defined.

### DIFF
--- a/fixtures/hasura/chinook/metadata/chinook.hml
+++ b/fixtures/hasura/chinook/metadata/chinook.hml
@@ -21,11 +21,25 @@ definition:
           comparison_operators:
             _eq:
               type: equal
+            _in:
+              type: custom
+              argument_type:
+                type: array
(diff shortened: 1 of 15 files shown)


fixtures/hasura/chinook/metadata/chinook.hml
```diff
@@ -21,11 +21,25 @@ definition:
           comparison_operators:
             _eq:
               type: equal
+            _in:
+              type: custom
```
daniel-chambers, line 25: This isn't quite right. `_in` should be using the standard `in` typed comparison operator: `ComparisonOperatorDefinition::In`, not a custom operator.

hallettj: Good point! I've updated the PR

daniel-chambers (approved):

PR #125, opened 2024-11-18
remove cloudflare resolver config

shahidhk:
having this will cause issues if cloudflare resolver is not reachable


hallettj (approved):
Good catch!

PR #126, opened 2024-11-19
support $project stage in native query pipeline type inference

hallettj:
This is work an an in-progress feature that is gated behind a feature flag, `native-query-subcommand`.

This change allows the native query configuration generator to infer the output type of [`$project`](https://www.mongodb.com/docs/manual/reference/operator/aggregation/project/) stages in pipelines. It also enables inference for parameters used in `$limit` and `$skip` stages.

`$project` is a complicated feature, but it's important to support it because it's widely used. It allows adding, removing, or modifying document fields in the document-processing pipeline. It has two modes: it can either remove fields, or it can select a subset of fields to keep while optionally adding more or changing values of kept fields. It also has dotted-path notation for easily manipulating nested structures.

### DIFF
--- a/crates/cli/src/native_query/pipeline/project_stage.rs
+++ b/crates/cli/src/native_query/pipeline/project_stage.rs
@@ -1,0 +1,348 @@
+use std::{
+    collections::{hash_map::Entry, HashMap},
+    str::FromStr as _,
+};
+
+use itertools::Itertools as _;
+use mongodb::bson::{Bson, Decimal128, Document};
+use mongodb_support::BsonScalarType;
+use ndc_models::{FieldName, ObjectTypeName};
+use nonempty::NonEmpty;
+
+use crate::native_query::{
+    aggregation_expression::infer_type_from_aggregation_expression,
+    error::{Error, Result},
+    pipeline_type_context::PipelineTypeContext,
+    type_constraint::{ObjectTypeConstraint, TypeConstraint},
+};
+
+enum Mode {
+    Exclusion,
+    Inclusion,
+}
+
+// $project has two distinct behaviors:
+//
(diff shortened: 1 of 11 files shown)


crates/cli/src/native_query/tests.rs
```diff
@@ -323,6 +324,121 @@ fn supports_various_aggregation_operators() -> googletest::Result<()> {
...
+                    ]),
+                    description: None
+                }
+            )
+        ]
+        .into(),
+    );
+
+    Ok(())
+}
```
hallettj, line 440: These tests demonstrates inference for `$project` in its two possible modes.

crates/cli/src/native_query/pipeline/project_stage.rs
```diff
@@ -0,0 +1,444 @@
...
+// Inclusion (replace) mode: if any value in the projection document specifies a field for
+// inclusion, replaces the value of an input field with a new value, adds a new field with a new
+// value, or removes a field with the special value `$$REMOVE` then output excludes input fields
+// that are not specified. The output is composed solely of fields specified in the projection
+// document, plus `_id` unless `_id` is specifically excluded. Values of `false` or `0` are not
+// allowed in this mode except to suppress `_id`.
+//
+// TODO: This implementation does not fully account for uses of $$REMOVE. It does correctly select
+// inclusion mode if $$REMOVE is used. A complete implementation would infer a nullable type for
+// a projection that conditionally resolves to $$REMOVE.
```
codedmart, line 39: Have you filed a ticket for this so we can come back to it at some point?

hallettj: I filed https://linear.app/hasura/issue/ENG-1365/infer-nullable-type-in-presence-of-dollardollarremove-in-dollarproject

PR #128, opened 2024-11-20
parse parameter type annotations when generating native query configuration

hallettj:
This is work an an in-progress feature that is gated behind a feature flag, `native-query-subcommand`.

Parses type annotations in placeholders that reference native query parameters. We already have an error message suggesting doing this, now the system actually reads these. For example,

```json
{
  "$match": {
    "imdb.rating": { "$gt": "{{ min_rating | int! }}" }
  }
}
```

The generated query for that configuration includes a parameter named `min_rating` with type `int`. Without the annotation it would have inferred `double`.

Type annotations are checked against the inferred type for the position so you will see errors if the annotated type is not compatible. Parameters are treated as contravariant so you can only provide a subtype of the inferred type - for example you can constrain a parameter type to be non-nullable even if it is in a nullable context. There are cases where the type checker cannot infer a type in which case annotations are necessary.

I know we already have a similar parser for type expressions in hml files. I thought it would be easier to write a new one specialized for this connector and for MongoDB scalar types since it's about 100 LOC.

Type expressions use GraphQL syntax to match types as seen in GraphQL and in hml. There is one addition: I invented a syntax for predicate types, `predicate<ObjectTypeName>`. The parser happens to be written so that if the angle brackets are absent the word `predicate` will be interpreted as an object type so we don't have a problem if a user want to use an object type named "predicate".

On the other hand this parser does not allow object type names that match MongoDB scalar type names.

While I was working on this I noticed some issues with unifying types in the presence of nullability, and with displaying errors when a type annotation can't be unified with inferred constraints for a parameter's context. So I included some fixes in those areas.

### DIFF
--- a/crates/cli/src/native_query/error.rs
+++ b/crates/cli/src/native_query/error.rs
@@ -78,9 +78,6 @@ pub enum Error {
     #[error("Error parsing a string in the aggregation pipeline: {0}")]
     UnableToParseReferenceShorthand(String),
 
-    #[error("Error parsing parameter type annotation: {0}")]
-    UnableToParseTypeAnnotation(String),
-
     #[error("Type inference is not currently implemented for the query document operator, {0}. Please file a bug report, and declare types for your native query by hand for the time being.")]
     UnknownMatchDocumentOperator(String),
 
--- a/crates/cli/src/native_query/error.rs
+++ b/crates/cli/src/native_query/error.rs
@@ -55,9 +55,12 @@ pub enum Error {
         field_name: FieldName,
     },
 
-    #[error("Type mismatch in {context}: {a:?} is not compatible with {b:?}")]
+    #[error("Type mismatch{}: {a:?} is not compatible with {b:?}", match context {
+        Some(context) => format!(" in {}", context),
+        None => String::new(),
+    })]
     TypeMismatch {
-        context: String,
+        context: Option<String>,
         a: TypeConstraint,
         b: TypeConstraint,
     },
@@ -114,7 +117,7 @@ fn unable_to_infer_types_message(
         for name in problem_parameter_types {
             message += &format!("- {name}\n");
         }
-        message += "\nTry adding type annotations of the form: {{parameter_name|[int!]!}}\n";
+        message += "\nTry adding type annotations of the form: {{ parameter_name | [int!]! }}\n";
     }
     if could_not_infer_return_type {
         message += "\nUnable to infer return type.";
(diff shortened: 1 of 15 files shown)


crates/cli/src/native_query/type_solver/simplify.rs
```diff
@@ -212,61 +268,16 @@ fn simplify_constraint_pair(
...
-                    object_type_name: b,
-                },
-            )),
+            })),
+            None => Ok(None),
         },
 
-        // TODO: We probably want a separate step that swaps ElementOf and FieldOf constraints with
-        // constraint of the targeted structure. We might do a similar thing with
-        // WithFieldOverrides.
```
codedmart, line 230: Follow up tickets?

crates/cli/src/native_query/type_solver/simplify.rs
```diff
@@ -574,4 +603,137 @@ mod tests {
...
+                    context: None,
+                    a: TypeConstraint::Scalar(BsonScalarType::Decimal),
+                    b: TypeConstraint::Scalar(BsonScalarType::Null),
+                }),
+            ])
+        );
+        Ok(())
+    }
+
+    // TODO:
```
codedmart, line 701: Unsupported right now?

codedmart (approved):

crates/cli/src/native_query/error.rs
```diff
@@ -114,7 +117,8 @@ fn unable_to_infer_types_message(
         for name in problem_parameter_types {
             message += &format!("- {name}\n");
         }
-        message += "\nTry adding type annotations of the form: {{parameter_name|[int!]!}}\n";
+        message += "\nTry adding type annotations of the form: {{ parameter_name | [int!]! }}\n";
+        message += "\nIf you added an annotation, and you are still seeing this error then the type you gave may not be compatible with the context where the parameter is used.\n";
```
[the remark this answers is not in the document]
hallettj: I need to do a docs pass on the whole native query DX feature, then I'll have something to link to

crates/cli/src/native_query/type_solver/simplify.rs
hallettj: I removed the TODO because I don't think we need to worry about this

crates/cli/src/native_query/type_solver/simplify.rs
hallettj: Yes - I wrote the test to see if it works, and it doesn't. I want to fix that later so I want to keep the test, but I don't want to commit a failing test.

PR #132, opened 2024-11-27
add issue template for native query dx

hallettj:


codedmart (commented):
Do we want this one as well as the one you have in #131?

hallettj:
Good point! I forgot I opened a redundant PR

PR #133, opened 2024-12-04
Remove volume from dockerfile

codedmart:
Setting the volumn in the dockerfile is causing issues during local development where the volumn gets created and then not updated when you reintrospect or change the configuration.json file.

- [x] TODO: Update Changelog file


hallettj (approved):
Ah, good find

PR #134, opened 2024-12-05
recreate supergraph fixture using latest cli version

hallettj:
Recreates the supergraph in `fixtures/hasura/` using ddn cli v2.15.0 with as little hand-editing of metadata as possible. This makes it much easier to continue updating metadata using the cli.

This PR mainly affects the integration test and local development testing setup. But since I needed to update the GraphQL Engine version used in integration tests I updated flake inputs. That leads to updating the openssl version. And since all of that leads to rebuilding stuff anyway I also updated the Rust toolchain to v1.83.0.

An issue that came up is that the hack I had tested to filter by array fields does not seem to work anymore. It looks like this is because of a change in the engine - the relevant test fails with the previous metadata fixture with the latest engine version. I've disabled the test. This is the test query:

```graphql
query {
  movies(where: { cast: { _eq: "Albert Austin" } }) {
    title
    cast
  }
}
```

Note that `cast` is an array of strings. It looks like we may need to wait for ndc-spec v0.2 to be able to do this properly.

The changes here include a lot of automatically-generated configuration in `fixtures/hasura/` so it's a big set of changes.


crates/integration-tests/src/tests/filtering.rs
```diff
@@ -66,24 +66,25 @@ async fn filters_by_comparisons_on_elements_of_array_field() -> anyhow::Result<(
...
-              }
-            }
-            "#
-        )
-        .run()
-        .await?
-    );
-    Ok(())
-}
+// This doesn't seem to work anymore in the latest engine version
```
danieljharvey, line 69: This can be re-enabled if you put the `CompatibilityConfig` date back I believe?

fixtures/hasura/app/connector/chinook/schema/Playlist.json
```diff
@@ -10,9 +10,7 @@
       "fields": {
         "Name": {
           "type": {
-            "nullable": {
```
danieljharvey, line 13: I am assuming this is expected?

danieljharvey (approved):

crates/integration-tests/src/tests/filtering.rs
hallettj: I tried setting `CompatibilityConfig` to an earlier date, and it didn't help so I'm removing this test. This functionality isn't supported until the next ndc-spec version anyway.

fixtures/hasura/app/connector/chinook/schema/Playlist.json
hallettj: Yeah that's nothing to worry about

PR #136, opened 2024-12-06
implement aggregations on nested fields

hallettj:
Turns on the nested field aggregation connector capability, and makes it work.

This also touches up the aggregation implementation generally. It uses the updated, unified system for referencing fields with paths. It also changes aggregation result types (besides count) to be nullable which avoids an error when aggregating over an empty document set.


crates/mongodb-agent-common/src/scalar_types_capabilities.rs
```diff
@@ -147,20 +145,23 @@ fn bson_to_named_type(bson_scalar_type: BsonScalarType) -> Type {
 
 pub fn aggregate_functions(
     scalar_type: BsonScalarType,
-) -> impl Iterator<Item = (AggregationFunction, BsonScalarType)> {
-    [(A::Count, S::Int)]
+) -> impl Iterator<Item = (AggregationFunction, Type)> {
+    let nullable_scalar_type = move || Type::Nullable {
+        underlying_type: Box::new(bson_to_named_type(scalar_type)),
+    };
```
codedmart, line 151: Not blocking, but is this useful in other areas as well?

hallettj: It could be slightly useful to have a method on `ndc_models::Type` that does this. That would be a change in the ndc-spec repo. In the connector code we already have `configuration::Type::make_nullable`.

PR #140, opened 2025-01-16
handle collection validators with object fields that do not list properties

hallettj:
If a collection validator species an property of type `object`, but does not specify a list of nested properties for that object then we will infer the `ExtendedJSON` type for that property. For a collection created with this set of options would have the type `ExtendedJSON` for its `reactions` field:

```json
{
  "validator": {
    "$jsonSchema": {
      "bsonType": "object",
      "properties": {
        "reactions": { "bsonType": "object" },
      }
    }
  }
}
```

If the validator specifies a map of nested properties, but that map is empty, then we interpret that as an empty object type.


crates/cli/src/introspection/validation_schema.rs
```diff
@@ -151,11 +150,14 @@ fn make_field_type(object_type_name: &str, prop_schema: &Property) -> (Vec<Objec
     let mut collected_otds: Vec<ObjectType> = vec![];
 
     match prop_schema {
+        Property::Object {
+            properties: None, ..
+        } => (vec![], Type::ExtendedJSON),
```
daniel-chambers, line 155: The meat of the change.

daniel-chambers (approved):
LGTM

PR #144, opened 2025-02-06
implement group by for single-column aggregates

hallettj:
Implements most of the functionality for the capability `query.aggregates.group_by`. There are still a couple of things to follow up on.

Counts are not implemented for group by queries yet. I'll follow up on those in [ENG-1568](https://linear.app/hasura/issue/ENG-1568/[mongodb]-implement-count-for-group-by). (Counts are implemented in #145 which can be merged after this PR is merged.)

There is a bug involving multiple references to the same relationship that should be resolved. I'll follow up in [ENG-1569](https://linear.app/hasura/issue/ENG-1569).

While working on this I removed the custom "count" aggregation - it is redundant, and I've been meaning to do that for a while. Users can use the standard count aggregations instead.

There is a change in here that explicitly converts aggregate result values for "average" and "sum" aggregations to the result types declared in the schema. This is necessary to avoid errors in response serialization for groups when aggregating over 128-bit decimal values. I applied the same type conversion for group and for root aggregates for consistency. This does mean there will be some loss of precision in those cases. But it also means we won't get back a JSON string in some cases, and a JSON number in others.


crates/mongodb-agent-common/src/query/groups.rs
```diff
@@ -0,0 +1,162 @@
...
+type Result<T> = std::result::Result<T, MongoAgentError>;
+
+// TODO: This function can be infallible once ENG-1562 is implemented.
+pub fn pipeline_for_groups(grouping: &Grouping) -> Result<Pipeline> {
+    let group_stage = Stage::Group {
+        key_expression: dimensions_to_expression(&grouping.dimensions).into(),
+        accumulators: accumulators_for_aggregates(&grouping.aggregates)?,
+    };
+
+    // TODO: ENG-1562 This implementation does not fully implement the
```
codedmart, line 26: Do we need to do this work at a specific time or this is behind a capability?

crates/mongodb-agent-common/src/query/groups.rs
```diff
@@ -0,0 +1,162 @@
...
+                    column_name,
+                    field_path.as_ref(),
+                ),
+            };
+            column_ref.into_aggregate_expression().into_bson()
+        })
+        .collect()
+}
+
+// TODO: This function can be infallible once counts are implemented
```
codedmart, line 77: Is there follow up work we need here?

crates/mongodb-agent-common/src/query/groups.rs
```diff
@@ -0,0 +1,162 @@
...
+fn accumulators_for_aggregates(
+    aggregates: &IndexMap<FieldName, Aggregate>,
+) -> Result<BTreeMap<String, Accumulator>> {
+    aggregates
+        .into_iter()
+        .map(|(name, aggregate)| Ok((name.to_string(), aggregate_to_accumulator(aggregate)?)))
+        .collect()
+}
+
+// TODO: This function can be infallible once counts are implemented
```
codedmart, line 87: Same question as above?

crates/mongodb-agent-common/src/query/response.rs
```diff
@@ -182,18 +207,68 @@ fn serialize_rows(
...
+                        dimension_definition.value_type(),
+                        dimension_value.clone(),
+                    )?)
+                })
+                .collect::<Result<_>>()?;
+
+            let aggregates = serialize_aggregates(mode, path, &grouping.aggregates, doc.into())?;
+
+            // TODO: This conversion step can be removed when the aggregates map key type is
+            // changed from String to FieldName
```
codedmart, line 245: Is there a follow up ticket for this one?

crates/ndc-query-plan/src/plan_for_query_request/plan_for_grouping.rs
```diff
@@ -0,0 +1,245 @@
...
+}
+
+fn plan_for_group_comparison_target<T: QueryContext>(
+    plan_state: &mut QueryPlanState<T>,
+    object_type: &plan::ObjectType<T::ScalarType>,
+    target: ndc::GroupComparisonTarget,
+) -> Result<plan::GroupComparisonTarget<T>> {
+    let plan_target = match target {
+        // TODO: Do we expect the target aggregate to correspond to one of the grouping aggregates?
+        // Or can it be independent?
```
codedmart, line 170: This seems like a question? Do we know the answer and can remove this?

crates/ndc-query-plan/src/plan_for_query_request/plan_for_grouping.rs
```diff
@@ -0,0 +1,245 @@
...
+    collection_object_type: &plan::ObjectType<T::ScalarType>,
+    target: ndc::GroupOrderByTarget,
+) -> Result<plan::GroupOrderByTarget<T>> {
+    match target {
+        ndc::GroupOrderByTarget::Dimension { index } => {
+            Ok(plan::GroupOrderByTarget::Dimension { index })
+        }
+        ndc::GroupOrderByTarget::Aggregate { aggregate } => {
+            // TODO: Do we expect the target aggregate to correspond to one of the grouping aggregates?
+            // Or can it be independent?
```
codedmart, line 237: This seems like a question? Do we know the answer and can remove this?

crates/ndc-query-plan/src/query_plan/requests.rs
```diff
@@ -0,0 +1,171 @@
...
+    pub variables: Option<Vec<VariableSet>>,
+
+    /// Types for values from the `variables` map as inferred by usages in the query request. It is
+    /// possible for the same variable to be used in multiple contexts with different types. This
+    /// map provides sets of all observed types.
+    ///
+    /// The observed type may be `None` if the type of a variable use could not be inferred.
+    pub variable_types: VariableTypes<T::ScalarType>,
+
+    // TODO: type for unrelated collection
```
codedmart, line 31: Do we need a ticket here to track this?

codedmart (approved):
Left a few comments that are not relevant to merging this. I mostly did reading and spot checking, but seems fine to me.

crates/mongodb-agent-common/src/query/groups.rs
hallettj: Oh right, that's behind the query.aggregates.group_by.order capability

crates/mongodb-agent-common/src/query/groups.rs
hallettj: I've followed up on this in #145

crates/mongodb-agent-common/src/query/response.rs
hallettj: Effectively yes. That change is in the latest ndc-spec, and I have a ticket for updating the spec https://linear.app/hasura/issue/ENG-1573/[mongodb]-update-to-latest-ndc-spec-revision

crates/ndc-query-plan/src/plan_for_query_request/plan_for_grouping.rs
hallettj: Good catch - I found the answer but forgot to remove the comment

crates/ndc-query-plan/src/query_plan/requests.rs
hallettj: Maybe - I don't remember where this would come up

PR #147, opened 2025-02-26
do not use $lookup for query with variables if only single variable set

hallettj:
This allows a `$vectorSearch` to be the target of a remote join if the join forwards exactly one variable set.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -3,6 +3,8 @@ on:
   push:
     branches:
       - main
+      - 'v*'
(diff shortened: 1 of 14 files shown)


hallettj:
In addition to MongoDB Atlas not allowing `$vectorSearch` in sub-pipelines it looks like `$vectorSearch` does not substitute variable references in its arguments. So even after I added a workaround for the sub-pipeline issue yesterday the remote join use case still didn't work. In particular we had a problem with the `queryVector` argument.

There is a way to fix the variable issue with some more work: in the case of a query request with exactly one variable set rewrite the request to inline all variable values. That's another way to get the sub-pipeline workaround so it would replace the changes currently in this PR.

PR #148, opened 2025-02-28
add uuid scalar type

hallettj:
UUIDs are stored in BSON using the `binData` type which we already support. But the `binData` type is not especially user friendly to work with in its generalized form, and UUIDs are a common use case. It is helpful to surface UUIDs as a first-class scalar type that serializes to a string.

The `binData` type has a numeric `subType` field to give a hint as to how the stored binary data should be interpreted. There are two subtypes for UUIDs, 3 & 4, but subtype 4 is indicated as the "current" representation for UUIDs. The `UUID()` constructor in `mongosh` produces `binData` values with subtype 4.

This change:
- adds a scalar type called `UUID`
- sets the NDC type representation for `UUID` to `String`
- converts `UUID` values to BSON as `binData` values with subtype 4 - inputs are parsed from string representations using functions provided by the BSON library that is bundled with the MongoDB Rust driver
- serializes `UUID` values to JSON as strings using provided BSON library functions - for example, `4ca4b7e7-6f6a-445b-b142-9d6d252d92bc`
- updates introspection to infer the type `UUID` when encountering fields with `binData` values with subtype 4
- introspection infers the type `BinData` instead for fields where subtype 4 occurs alongside `binData` values with other subtypes

This allows for more user-friendly queries involving UUIDs. For example filtering by UUID changed from this:

```gql
id: { _eq: { base64: "TKS3529qRFuxQp1tJS2SvA==" subType: "04" } }
```

and now looks like this:

```gql
id: { _eq: "4ca4b7e7-6f6a-445b-b142-9d6d252d92bc" }
```

### DIFF
--- a/crates/cli/src/introspection/type_unification.rs
+++ b/crates/cli/src/introspection/type_unification.rs
@@ -48,13 +48,9 @@ pub fn unify_type(type_a: Type, type_b: Type) -> Type {
         // Scalar types unify if they are the same type, or if one is a superset of the other.
         // If they are diffferent then the union is ExtendedJSON.
         (Type::Scalar(scalar_a), Type::Scalar(scalar_b)) => {
-            if scalar_a == scalar_b || is_supertype(&scalar_a, &scalar_b) {
-                Type::Scalar(scalar_a)
-            } else if is_supertype(&scalar_b, &scalar_a) {
-                Type::Scalar(scalar_b)
-            } else {
-                Type::ExtendedJSON
-            }
+            BsonScalarType::common_supertype(scalar_a, scalar_b)
+                .map(Type::Scalar)
+                .unwrap_or(Type::ExtendedJSON)
         }
 
         // Object types unify if they have the same name.
@@ -190,20 +186,6 @@ pub fn unify_object_types(
     );
 
     merged_type_map.into_values().collect()
-}
-
-/// True iff we consider a to be a supertype of b.
-///
-/// Note that if you add more supertypes here then it is important to also update the custom
-/// equality check in our tests in mongodb_agent_common::query::serialization::tests. Equality
-/// needs to be transitive over supertypes, so for example if we have,
-///
-/// (Double, Int), (Decimal, Double)
-///
-/// then in addition to comparing ints to doubles, and doubles to decimals, we also need to compare
-/// decimals to ints.
-pub fn is_supertype(a: &BsonScalarType, b: &BsonScalarType) -> bool {
-    matches!((a, b), (Double, Int))
 }
 
 #[cfg(test)]
--- a/crates/mongodb-agent-common/src/query/serialization/json_to_bson.rs
+++ b/crates/mongodb-agent-common/src/query/serialization/json_to_bson.rs
@@ -71,11 +71,12 @@ pub fn json_to_bson(expected_type: &Type, value: Value) -> Result<Bson> {
 
 /// Works like json_to_bson, but only converts BSON scalar types.
 pub fn json_to_bson_scalar(expected_type: BsonScalarType, value: Value) -> Result<Bson> {
+    use BsonScalarType as S;
     let result = match expected_type {
-        BsonScalarType::Double => Bson::Double(deserialize(expected_type, value)?),
-        BsonScalarType::Int => Bson::Int32(deserialize(expected_type, value)?),
-        BsonScalarType::Long => convert_long(&from_string(expected_type, value)?)?,
-        BsonScalarType::Decimal => Bson::Decimal128(
+        S::Double => Bson::Double(deserialize(expected_type, value)?),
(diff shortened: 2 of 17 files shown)


crates/cli/src/introspection/type_unification.rs
```diff
@@ -192,20 +188,6 @@ pub fn unify_object_types(
...
-/// equality check in our tests in mongodb_agent_common::query::serialization::tests. Equality
-/// needs to be transitive over supertypes, so for example if we have,
-///
-/// (Double, Int), (Decimal, Double)
-///
-/// then in addition to comparing ints to doubles, and doubles to decimals, we also need to compare
-/// decimals to ints.
-pub fn is_supertype(a: &BsonScalarType, b: &BsonScalarType) -> bool {
-    matches!((a, b), (Double, Int))
-}
```
hallettj, line 207: I moved this to a method on `BsonScalarType`

crates/mongodb-agent-common/src/query/serialization/json_to_bson.rs
```diff
@@ -84,37 +85,34 @@ pub fn json_to_bson_scalar(expected_type: BsonScalarType, value: Value) -> Resul
...
-        BsonScalarType::BinData => {
-            deserialize::<json_formats::BinData>(expected_type, value)?.into()
-        }
-        BsonScalarType::ObjectId => Bson::ObjectId(deserialize(expected_type, value)?),
-        BsonScalarType::Bool => match value {
+        S::String => Bson::String(deserialize(expected_type, value)?),
+        S::Date => convert_date(&from_string(expected_type, value)?)?,
+        S::Timestamp => deserialize::<json_formats::Timestamp>(expected_type, value)?.into(),
+        S::BinData => deserialize::<json_formats::BinData>(expected_type, value)?.into(),
+        S::UUID => convert_uuid(&from_string(expected_type, value)?)?,
```
hallettj, line 92: I'm sorry about the churn in this function. This is the only added line. The logic in other lines is unchanged.

crates/mongodb-support/src/bson_type.rs
```diff
@@ -80,21 +80,7 @@ impl<'de> Deserialize<'de> for BsonType {
...
-    PartialOrd,
-    Ord,
-    Sequence,
-    Serialize,
-    Deserialize,
-    JsonSchema,
-)]
-#[serde(try_from = "BsonType", rename_all = "camelCase")]
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash, PartialOrd, Ord, Sequence, JsonSchema)]
 pub enum BsonScalarType {
```
hallettj, line 84: We serialize BSON scalar types using the names in `bson_name()`. But so for this has been implicit - the serde serialization and the `bson_name` function didn't reference each other. The `camelCase` renaming just happened to match the strings in `bson_name`.

I changed this now because serde camel-cases `UUID` to `uUID` which I think is wrong. I replaced the derived `Serialize` and `Deserialize` implementations with hand-written implementations that call `bson_name` and `from_bson_name`.

I also removed the `try_from = "BsonType"` part. That has a hack for parsing collection validators. In the `schema::Property::Scalar` variant we declare the `bson_type` field as `BsonScalarType`, but actually it can be an array type. So we need to parse that field as a `BsonType`, and attempt to convert it to a scalar type. Instead of doing that for all `BsonScalarType` deserializations I moved that logic into the `schema` module.

crates/mongodb-support/src/bson_type.rs
```diff
@@ -259,7 +278,60 @@ impl BsonScalarType {
...
+            } else {
+                match (a, b) {
+                    (S::Double, S::Int) => Some(S::Double),
+                    _ => None,
+                }
+            }
         }
+        helper(a, b).or_else(|| helper(b, a))
+    }
+}
```
hallettj, line 317: `common_supertype` is more general than `is_supertype`: it can provide a common ancestor type given two types that don't overlap, or don't fully overlap with each other. This is a bit of future-proofing in case we add more binary subtype scalar types. For example if a field is found to have `UUID` and also `Vector` values we want to unify those to `BinData`.

crates/mongodb-support/src/bson_type.rs
```diff
@@ -259,7 +278,60 @@ impl BsonScalarType {
             S::Undefined => true,
             S::DbPointer => true,
             S::Symbol => true,
+            S::UUID => true,
+        }
+    }
+
+    /// True iff we consider a to be a supertype of b.
```
codedmart, line 285: Nit typo

codedmart (approved):

crates/mongodb-support/src/bson_type.rs
hallettj: That's not a typo - "iff" is short for "if and only if"

PR #152, opened 2025-03-07
introspection updates existing schemas to add fields only

hallettj:
Previously running introspection would not update existing schema definitions, it would only add definitions for newly-added collections. This PR changes that behavior to make conservative changes to existing definitions: added fields, either top-level or nested, will be added to existing schema definitions. Types for fields that are already configured will not be changed. Fields that appear to have been added to collections will not be removed from configurations.

I included output that shows any detected database changes that were not applied to configuration. Here's what that looks like:

```
Warning: introspection detected some changes to to database thate were **not** applied to existing
schema configurations. To avoid accidental breaking changes the introspection system is
conservative about what changes are applied automatically.

To apply changes delete the schema configuration files you want updated, and run introspection
again; or edit the files directly.

These database changes were **not** applied:

app/connector/test_cases/schema/uuids.json:
 {
   objectTypes: {
     uuids: {
       fields: {
         uuid: {
           type: {
-            scalar: "uuid"
+            scalar: "binData"
           }
         }
       }
     }
   }
 }
```


crates/cli/src/lib.rs
```diff
@@ -91,22 +93,42 @@ async fn update(
...
         database,
-        &existing_schemas,
+        existing_schemas,
     )
     .await?;
-    configuration::write_schema_directory(&context.path, schemas_from_sampling).await
+    configuration::write_schema_directory(&context.path, schemas_from_sampling).await?;
+
+    if !ignored_changes.is_empty() {
+        eprintln!("Warning: introspection detected some changes to to database thate were **not** applied to existing
```
danieljharvey, line 117:
```suggestion
        eprintln!("Warning: introspection detected some changes to to database that were **not** applied to existing
```

crates/cli/src/lib.rs
```diff
@@ -91,22 +93,42 @@ async fn update(
...
+schema configurations. To avoid accidental breaking changes the introspection system is
+conservative about what changes are applied automatically.");
+        eprintln!();
```
danieljharvey, line 120: This is all extremely explicit about what is going on, which is excellent.

PR #160, opened 2025-04-16
skip system and unsample-able collections in introspection

hallettj:
We have had users unable to run introspection because sampling fails on an automatically-generated collection called `system.views`. This collection is generated if the database has any views. On some deployments attempts to run aggregate (which is what introspection sampling uses) fail with a permissions error. We've seen this come up in MongoDB v6 running on Atlas.

Collections prefixed with `system.` are reserved for internal use. We shouldn't include them in introspection. https://www.mongodb.com/docs/manual/reference/system-collections/#synopsis

This PR makes two changes:
- skip collections whose names begin with `system.`, and log a warning
- when sampling a collection fails for any reason skip that collection and log a warning instead of failing the entire introspection process

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,11 +10,13 @@ This changelog documents the changes between release versions.
 
 ### Fixed
 
+- Database introspection no longer fails if any individual collection cannot be sampled ([#160](https://github.com/hasura/ndc-mongodb/pull/160))
+
 ## [1.7.1] - 2025-03-12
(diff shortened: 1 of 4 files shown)


hallettj:
I wrote an integration test for this, but since our integration test database runs without any security measures in place it was taking too long for me to reproduce the error. So no automated test. Sorry!

daniel-chambers (approved):

PR #162, opened 2025-04-25
add option to skip rows on response type mismatch

hallettj:
When sending response data for a query if we encounter a value that does not match the type declared in the connector schema the default behavior is to respond with an error. That prevents the user from getting any data. This change adds an option to silently skip rows that contain type mismatches so that the user can get a partial set of result data.


codedmart (approved):
:tada:

PR #166, opened 2025-05-08
Change generated query logs to info level

danieljharvey:
These are helpful for users.


hallettj (approved):
The reason this was at debug is because the pipeline will include user-entered query values, like values used in filter comparisons. I thought there might be some possible scenario where there might be sensitive data in there. But maybe that is too paranoid; and being able to see pipelines does seem very useful.

danieljharvey:
> The reason this was at debug is because the pipeline will include user-entered query values, like values used in filter comparisons. I thought there might be some possible scenario where there might be sensitive data in there. But maybe that is too paranoid; and being able to see pipelines does seem very useful.

That's fair. Perhaps we should make this opt-in with a particular env var. Will have a think.

PR #169, opened 2025-06-11
Add aws, azure, gcp auth to features

codedmart:


Cargo.toml
```diff
@@ -25,13 +25,14 @@ indexmap = { version = "2", features = [
   "serde",
 ] } # should match the version that ndc-models uses
 itertools = "^0.12.1"
-mongodb = { version = "^3.1.0", features = ["tracing-unstable"] }
+mongodb = { version = "^3.1.0", features = ["tracing-unstable", "aws-auth", "azure-kms", "azure-oidc", "gcp-oidc"] }
 schemars = "^0.8.12"
 serde = { version = "1", features = ["derive"] }
 serde_json = { version = "1.0", features = ["preserve_order", "raw_value"] }
 ref-cast = "1.0.23"
+reqwest = "0.12.20"
```
hallettj, line 33: What's the purpose of adding the reqwest version constraint here? After this change we're going to have two copies of reqwest anyway. The added mongodb driver features specify an exact dependency on this version anyway. And we also have ndc-sdk and opentelemetry libs specifying an exact dependency on reqwest 0.11.

PR #187, opened 2026-05-18
Add env-loader image variant for external secrets support

dliub:
## Summary

- Adds `Dockerfile.env-loader` for an Alpine-based image that bundles `jq`, the statically-linked `mongodb-connector` binary, and a small wrapper script. The wrapper reads every JSON file in `/secrets/` and exports each key/value pair as an environment variable before exec'ing the connector.
- Adds `scripts/env-loader.sh` (the wrapper).
- Adds a `docker-env-loader` job to `.github/workflows/deploy.yml` that consumes the binaries produced by `build-connector-binaries`, derives the same tag set as `scripts/publish-docker-image.nix`, and pushes a multi-arch image (`linux/amd64` + `linux/arm64`) to `ghcr.io/hasura/ndc-mongodb` with a `-env-loader` suffix on each tag (e.g. `v2.0.1-env-loader`, `dev-main-env-loader`).

## Why

The Nix-built connector image has no shell or package manager, so `jq` and the wrapper can't be layered on top of it directly. The cleanest path is to build a separate Alpine-based image that reuses the same statically-linked binary CI already produces, and publish it as a sibling tag. This is what [ddn-helm-charts PR #58](https://github.com/hasura/ddn-helm-charts/pull/58) expects for its external-secrets integration.

See PromptQL thread: https://prompt.ql.app/project/hasuraql/promptql-playground/thread/f13889a5-3ff7-4b3f-972f-f883e8994310

## Test plan

- [ ] CI publishes `ghcr.io/hasura/ndc-mongodb:<tag>-env-loader` alongside the existing `<tag>` (both branch and `v*` tag refs).
- [ ] `docker run --rm -v $(pwd)/secret.json:/secrets/secret.json ghcr.io/hasura/ndc-mongodb:<tag>-env-loader env` shows the JSON keys exported as env vars.
- [ ] Multi-arch image works on both `linux/amd64` and `linux/arm64`.
- [ ] Connector starts and serves on `:8080` when given a valid `MONGODB_DATABASE_URI` via the secret.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -78,6 +78,95 @@ jobs:
         path: release
         if-no-files-found: error
 
+  # Builds and publishes the `-env-loader` variant of the connector image.
+  # Uses the statically-linked binaries produced above + Alpine + jq so that
+  # the Helm chart's external-secrets integration can mount JSON secrets at
+  # /secrets/ and have them exported as env vars before the connector starts.
+  docker-env-loader:
+    name: deploy::docker (env-loader)
+    needs: build-connector-binaries
+    runs-on: ubuntu-24.04
+    permissions:
+      contents: read
+      packages: write
+    steps:
+      - name: Checkout 🛎️
+        uses: actions/checkout@v4
+
+      - name: Download connector binaries
+        uses: actions/download-artifact@v4
+        with:
+          path: release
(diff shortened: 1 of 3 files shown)


danieljharvey (approved):
Need to get static linking working in the other repos, makes this a lot simpler.

PR #190, opened 2026-07-24
feat(relational): compose collection-representation native queries via Relation::From

dliub:
## What & why

PromptQL's Virtual SQL Layer calls MongoDB Models/TVFs (including Atlas
`$search` / `$searchMeta` / `$vectorSearch` pipelines) through the **relational**
NDC protocol (`Relation::From`). Today the relational pipeline builder treats
`From.collection` as a *physical* MongoDB collection: it rejects non-empty
`From.arguments`, never resolves a configured native query, never interpolates
arguments, and never prepends the configured native pipeline. Argument-free
native queries are also incorrect in relational mode because the configured
pipeline is omitted.

This PR makes existing **collection-representation** native queries composable
through `Relation::From`, reusing the classic NDC native-query machinery
(`json_to_bson` typing/validation and `interpolated_command` substitution).

## Behavior

1. Resolve `From.collection` against physical collections **and** configured
   collection-representation native queries.
2. Validate and bind `From.arguments` (unknown / missing / null / type-invalid),
   reusing the classic BSON typing and `{{ argument }}` interpolation.
3. Select `input_collection` when configured, otherwise run a database-level
   `db.aggregate` (native queries with no input collection).
4. Keep the configured native pipeline as an **immutable source prefix**.
5. Append SQL-generated filter / sort / paginate / project / join / aggregate
   stages after that prefix.
6. Never inject the relational early-match `$match` before the native prefix, so
   `$search` / `$searchMeta` / `$vectorSearch` remain stage zero.
7. Reject function-representation native queries used as relational tables.
8. Preserve existing physical-collection relational behavior.

## Design

- `relational/native_query.rs` (new): native-query resolution, argument
  validation/binding (`json_to_bson` + `interpolated_command`), and prefix
  materialization.
- `relational/pipeline_builder.rs`: `build_from` resolves native queries;
  `PipelineContext` tracks the physical `target_collection`; early-match is
  skipped for native-query roots.
- `relational/types.rs`: `RelationalPipelineResult` gains `target_collection`
  (`Some` → `db.<name>.aggregate`, `None` → database-level `db.aggregate`).
- `relational/execute.rs`: chooses collection vs database-level aggregation; the
  two cursor types are unified with `futures::future::Either` (preserves `Send`).
- `relational/error.rs`: new error variants for the cases above.

See `DESIGN.md` in the task workspace for the full rationale.

## Tests

Focused tests in `crates/mongodb-agent-common/src/relational/tests/native_query_tests.rs`:

- TVF invocation with typed scalar arguments; representative BSON scalar types
  (String, Int, Long, Double, Bool, ObjectId, Date).
- Argument-free collection native query.
- Missing / unknown / null / type-invalid arguments.
- `$search` stays stage zero while generated WHERE / ORDER BY / LIMIT / OFFSET
  land after the native prefix.
- Native queries with (`input_collection`) and without (database-level) an input
  collection.
- ObjectId string→ObjectId coercion; relational-vs-classic interpolation parity.
- Rejection of function-representation native queries.
- Physical-collection regression (targeting + argument rejection).

Full crate suite: `286 passed; 0 failed`. `cargo clippy` clean.

## Scope / follow-ups

- **ndc-mongodb only.** v3-engine is untouched; the separate
  `ModelSource.argument_mappings` work is a follow-up.
- Not attempted (kept as distinct native-query patterns): rewriting SQL
  predicates into `$search.compound.filter`; rewriting `COUNT(1)` over `$search`
  rows into `$searchMeta`.
- `RelationalLiteral` cannot express arrays/objects, so composite list/object
  arguments fall back to classic NDC.
- A native query **without** `input_collection` used as a join/union source (no
  physical `from`) remains unsupported; root-position use is fully supported.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/crates/mongodb-agent-common/src/relational/error.rs
+++ b/crates/mongodb-agent-common/src/relational/error.rs
@@ -40,4 +40,37 @@ pub enum RelationalError {
     /// Expression nesting is too deep.
     #[error("Expression nesting too deep (limit: 512)")]
     ExpressionTooDeep,
+
+    /// A function-representation native query was used as a relational table.
+    #[error("Native query \"{0}\" is represented as a function and cannot be used as a relational table; only collection-representation native queries are supported")]
+    FunctionRepresentationNotSupported(String),
+
+    /// An argument was supplied that the native query does not declare.
+    #[error("Unknown argument \"{argument}\" supplied to native query \"{native_query}\"")]
+    UnknownArgument {
+        native_query: String,
+        argument: String,
+    },
+
+    /// A required native-query argument was not supplied.
+    #[error("Missing argument \"{argument}\" for native query \"{native_query}\"")]
+    MissingArgument {
+        native_query: String,
+        argument: String,
+    },
+
+    /// An argument value could not be validated/bound to the declared argument type.
+    #[error("Invalid value for argument \"{argument}\" of native query \"{native_query}\": {message}")]
+    ArgumentBindingError {
+        native_query: String,
+        argument: String,
+        message: String,
+    },
+
+    /// Interpolating a native query's configured pipeline failed.
+    #[error("Failed to interpolate pipeline for native query \"{native_query}\": {message}")]
+    InterpolationError {
+        native_query: String,
+        message: String,
+    },
 }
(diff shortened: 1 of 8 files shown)


dliub:
Intended assignee: @manasag (could not set the assignee field automatically — the authoring account lacks assign permission on this repo; a maintainer can assign).

